PR #10, opened 2020-11-23
Word category normalizer

MarinPostma:


src/normalizer/token_classifier.rs
```diff
@@ -0,0 +1,91 @@
...
+        } else if word.chars().all(|c| {
+            is_hard_separator = self.hard_separators.contains(&c);
+            self.soft_separators.contains(&c) || is_hard_separator
+        }) {
+            if is_hard_separator {
+                token.kind = TokenKind::Separator(SeparatorKind::Hard);
+            } else {
+                token.kind = TokenKind::Separator(SeparatorKind::Soft);
+            }
+            token
```
LegendreM, line 36: this seems a bit overengineered, suggesting that the tokenizer could return a multicharacter separator.
Moreover, it is no consistent with stop words,  we don't check if any stop word in the list is a subset of the token.
Moreover, if I create a tokenizer which detects smiley like `;-)` as one token, it would be treated as a separator by this normalizer.
I would just compare the provided token with the list of separators.

src/tokenizer.rs
```diff
@@ -1,16 +1,42 @@
...
+        Self {
+            pre_processor: Box::new(IdentityPreProcessor),
+            tokenizer: Box::new(UnicodeSegmenter),
+            normalizer: Box::new(IdentityNormalizer),
+        }
+    }
+}
+
+impl Pipeline {
+    fn set_tokenizer(mut self, tokenizer: impl InternalTokenizer + 'static) -> Self {
```
LegendreM, line 30: `set_tokenizer` VS `with_tokenizer` ?

src/tokenizer.rs
```diff
@@ -1,16 +1,42 @@
...
-pub type Pipeline = (Box<dyn PreProcessor + 'static>, Box<dyn InternalTokenizer + 'static>, Box<dyn Normalizer + 'static>);
+static DEFAULT_PIPELINE: Lazy<Pipeline> = Lazy::new(|| Pipeline::default());
 
-static DEFAULT_PIPELINE: Lazy<Pipeline> = Lazy::new(||
-    (Box::new(IdentityPreProcessor), Box::new(UnicodeSegmenter), Box::new(IdentityNormalizer)));
+pub struct Pipeline {
+    pre_processor: Box<dyn PreProcessor + 'static>,
+    tokenizer: Box<dyn InternalTokenizer + 'static>,
+    normalizer: Box<dyn Normalizer + 'static>,
+}
```
LegendreM, line 17: this is much clearer as a struct 👍

LegendreM (changes requested):

PR #14, opened 2020-11-26
replace HahSet with fst::Set in classifier

MarinPostma:


LegendreM (approved):

src/analyzer.rs
```diff
@@ -118,35 +122,35 @@ impl Default for AnalyzerConfig {
...
-pub struct Analyzer {
-    config: AnalyzerConfig,
+pub struct Analyzer<'a, A> {
+    config: AnalyzerConfig<'a, A>,
 }
 
-pub struct AnalyzedText<'a> {
+pub struct AnalyzedText<'a, A>
+where
+    A: AsRef<[u8]>
```
Kerollmops, line 140: Not sure but I think you don't need this bound here, only on implementations.

src/analyzer.rs
```diff
@@ -235,7 +244,8 @@ mod test {
 
     #[test]
     fn test_simple_latin() {
-        let analyzer = Analyzer::new(AnalyzerConfig::default());
+        let stop_words = Set::from_iter([""].iter()).unwrap();
```
Kerollmops, line 247: There is a default impl, it's usually better than doing this by hand, here you create one empty key, do you really want that?

Kerollmops: It you wanted an empty iterator a `None as Option<&str>` would have been better.

src/token_classifier.rs
```diff
@@ -1,17 +1,22 @@
...
 
 use crate::{Token, TokenKind};
 use crate::token::SeparatorKind;
 
 #[derive(Debug,  Clone)]
-pub struct TokenClassifier<'a> {
-    stop_words: &'a HashSet<String>,
+pub struct TokenClassifier<'a, A>
+where
+    A: AsRef<[u8]>
```
Kerollmops, line 10: Same here and everywhere else, no need for this bound. :)

MarinPostma: I will look at that! Thanks mops 😊

src/analyzer.rs
MarinPostma: thanks, I was not sure how to deal with that :)

PR #17, opened 2020-11-27
Small improvements to code quality

MarinPostma:
make some changes suggested by @kerollmops

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -166,10 +166,7 @@ where
     }
 }
 
-impl<'a, A> Analyzer<'a, A>
-where
-    A: AsRef<[u8]>
-{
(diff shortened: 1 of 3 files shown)


Kerollmops (changes requested):
There are other places where the `A: AsRef<[u8]>` bound is not needed, especially in the impl of the `AnalyzedText` struct.

MarinPostma:
@Kerollmops `AnalyzedText` tokens methods makes a call to the classify method of `TokenClassifier`, which itself calls `contains` on `Set` which is bound on `AsRef<u8>`. The best I could do is take out the `new` method on `TokenClassifier` to an impl without the trait bound, but I don't really see a reason for that here.

LegendreM (approved):

PR #23, opened 2020-12-16
Use hmm on jieba

LegendreM:
Use HMM on Jieba which enhance the tokenization with a small performance impact on Chinese scripts

HMM vs No-HMM time (sec)


MarinPostma (approved):
Lgtm

PR #31, opened 2021-03-25
make stop_words optional

irevoire:
This feature would allow us to stop creating empty fst (which are slow to create) on milli.
See https://github.com/meilisearch/milli/issues/111

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -116,15 +116,11 @@ where
         self.stop_words = Some(stop_words);
         self
     }
-
-    pub fn new(pipeline_map: HashMap<(Script, Language), Pipeline>, stop_words: Option<&'a Set<A>>) -> Self {
-        Self { pipeline_map, stop_words }
-    }
 }
 
 impl AnalyzerConfig<'_>
 {
-    pub fn new_with_pipeline(pipeline_map: HashMap<(Script, Language), Pipeline>) -> Self {
+    pub fn new(pipeline_map: HashMap<(Script, Language), Pipeline>) -> Self {
         Self { pipeline_map, stop_words: None }
     }
 }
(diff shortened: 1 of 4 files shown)


src/analyzer.rs
```diff
@@ -101,18 +101,37 @@ make_script! {
...
-    pub fn default_with_stopwords(stop_words: &'a Set<A>) -> Self {
+    pub fn with_stopwords(&mut self, stop_words: &'a Set<A>) -> &mut Self {
+        self.stop_words = Some(stop_words);
+        self
+    }
+
+    pub fn new(pipeline_map: HashMap<(Script, Language), Pipeline>, stop_words: Option<&'a Set<A>>) -> Self {
+        Self { pipeline_map, stop_words }
+    }
+}
```
LegendreM, line 123: because you created `with_stopwords` methods, we may remove the `stop_words` parameter here and set `None` as the default value

src/analyzer.rs
```diff
@@ -101,18 +101,37 @@ make_script! {
...
+
+impl AnalyzerConfig<'_>
+{
+    pub fn new_with_pipeline(pipeline_map: HashMap<(Script, Language), Pipeline>) -> Self {
+        Self { pipeline_map, stop_words: None }
+    }
```
LegendreM, line 129: This could be the `new` method of `AnalyzerConfig`, and if I want to add `stop_words` I'll call `with_stopwords(...)` method

LegendreM (changes requested):

src/analyzer.rs
irevoire: Done thanks :+1:

LegendreM (approved):

LegendreM (approved):

LegendreM:
bors try

PR #39, opened 2021-04-14
fix(separators): treat cyrillic chars as non-separators

shekhirin:
Resolves https://github.com/meilisearch/Tokenizer/issues/38

The reason why it happened is because cyrillic letter `ь` in Russian is a way to soften the previous letter (e.g. there's a word `Альфа` which has an English transcription `Al'fa`, where `'` means that softening). That `'` was treated as a separator after deunicoding.


src/token_classifier.rs
```diff
@@ -57,6 +57,7 @@ where
 
 fn classify_separator(c: char) -> Option<SeparatorKind> {
     match deunicode_char(c)?.chars().next()? {
+        _ if matches!(c, 'а'..='я' | 'А'..='Я') => None, // prevent deunicoding cyrillic chars (e.g. ь -> ' is incorrect)
```
MarinPostma, line 60: can you add separate ranges for cyrillic? According to the [unicode table](https://unicode-table.com/fr/#ipa-extensions) there are a bunch of character in this table that may be considered as separator.

MarinPostma (changes requested):

src/token_classifier.rs
shekhirin: hm, actually the cyrillic part of the table is only that one
image
which I match by `'а'..='я' | 'А'..='Я'`.

I think that's enough and we can use previous deunicode way for detecting separators?

MarinPostma: uhm, so this A as a different UTF8 code as the the latin A, very confusing... Could you use hex range instead, and add a comment to tell that this correspond to the cyrillic range? This will avoid any confusion in the future :)

shekhirin: haha yeah sure. Fixed.

src/token_classifier.rs
```diff
@@ -57,6 +57,8 @@ where
 
 fn classify_separator(c: char) -> Option<SeparatorKind> {
     match deunicode_char(c)?.chars().next()? {
+        // Prevent deunicoding cyrillic chars (e.g. ь -> ' is incorrect)
+        _ if matches!(c, '\u{0410}'..='\u{044f}') => None, // russian cyrillic letters [а-яА-Я]
```
MarinPostma, line 61:
```suggestion
        _ if ('\u{0410}'..='\u{044f}').contains(c) => None, // russian cyrillic letters [а-яА-Я]
```

MarinPostma (changes requested):
last small change and we go :)

src/token_classifier.rs
shekhirin: yep, that looks better. Fixed.

MarinPostma (approved):
thanks!!

bors merge

PR #43, opened 2021-04-15
Implement FromIterator for TokenStream

shekhirin:
Need for https://github.com/meilisearch/milli/pull/160

### DIFF
--- a/src/tokenizer/mod.rs
+++ b/src/tokenizer/mod.rs
@@ -8,6 +8,7 @@ pub use legacy_meilisearch::LegacyMeilisearch;
 
 use crate::Token;
 use crate::processors::ProcessedText;
+use std::iter::FromIterator;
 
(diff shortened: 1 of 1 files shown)


LegendreM:
Hello @shekhirin!
I don't understand the purpose of your implementation.
For me, the only way to create a `TokenStream` with `from_iterator` is from a `TokenStream`,
meaning that we consume a `TokenStream` to create a `TokenStream` 🤔.
Am I wrong?

shekhirin:
@LegendreM actually it was supposed to used in https://github.com/meilisearch/milli/pull/160 but as you said it's not needed and there's a better way to do so with which I agree.
However, I intended to use this `FromIterator` to construct a `TokenStream` from any iterable of `Token` which may not be an existing `TokenSteam`.

PR #44, opened 2021-04-22
Fix non-breaking space separator

shekhirin:
Resolves https://github.com/meilisearch/Tokenizer/issues/38


LegendreM:
Hello @shekhirin, welcome again 😃 , I think your PR can be merged like this.
However, I'm thinking about adding a new kind of separator (NonBreaking) which will be not indexed in the DB but will not split the word 🤔.
My approach will consider similar words, containing or not a non-breaking-separator, as the same word.
For example: 
- `SEG\u{00a0}WIC1` would be considered as the same word as `SEGWIC1`
- `испытательные` would be considered as the same word as  `испытателные`

do you think it's relevant to do it like this?

Thanks again for your contribution, enhancing the Cyrillic support in meilisearch! 🥇

shekhirin:
@LegendreM looks good to me. I've implemented the idea in the latest commit, could you check it please?

src/token_classifier.rs
```diff
@@ -56,9 +56,14 @@ where
...
+    match c {
+        // cyrillic "ь" shouldn't be deunicoded into "'" which is a separator
+        // non-breaking space shouldn't be converted into a whitespace which is a separator
+        'ь' | '\u{00a0}' => return Some(SeparatorKind::NonBreaking),
+        _ => ()
+    };
+
     match deunicode_char(c)?.chars().next()? {
-        // Prevent deunicoding cyrillic chars (e.g. ь -> ' is incorrect)
-        _ if ('\u{0410}'..='\u{044f}').contains(&c) => None, // russian cyrillic letters [а-яА-Я]
```
LegendreM, line 61: Why don't we support `Cyrillic letters [а-яА-Я]` anymore?

LegendreM (changes requested):
Hello @shekhirin, I think you should roll back your last commit 😅, I was just asking a question and didn't request any changes.
Moreover, I don't know if adding a NonBreakingSeparator is the best solution to handle this, for example, we could remove non-breaking-separators during the normalization process.

shekhirin:
@LegendreM ah ok. I thought you proposed a solution to implement 😬 
Reverted the last commit.

LegendreM:
Thank you for your contribution @shekhirin!

PR #59, opened 2021-11-04
Add mapping between bytes in original word and normalized word

Samyak2:
# Pull Request

## What does this PR do?

Towards fixing meilisearch/tokenizer#54

 - Added a `char_map` vector in Token
 - It stores the number of bytes in the normalized word for each grapheme cluster (character) in the original word
 - To use it in a highlighter, given the number of bytes to highlight, use the `num_graphemes_from_bytes` fn to get the number of chars to highlight.
    - Though, note that the meilisearch highlighter only gets the raw string (&str) instead of the token struct. Hence it will need major changes in the highlighter, or an alternate solution here.
 - It must also be noted that this adds a significant overhead to tokenization as a new vector is stored for *every* token.
    - and `deunicode` has to be run twice on the same token - once for each grapheme cluster and once for the entire word.

## PR checklist
- [x] Does this PR fix an existing issue?

This is my first Rust PR :smile:, please let me know if something's not right

### DIFF
--- a/src/analyzer.rs
+++ b/src/analyzer.rs
@@ -484,4 +484,29 @@ mod test {
         let analyzed: Vec<_> = analyzed.tokens().map(|token| token.word).collect();
         assert_eq!(analyzed, ["小", "化妆包"]);
     }
+
+    #[test]
+    fn test_grapheme_char_map() {
+        let analyzer = Analyzer::new(AnalyzerConfig::<Vec<u8>>::default());
+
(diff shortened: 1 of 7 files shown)


ManyTheFish:
Can you add some tests concerning your changes? Please. 😊

ManyTheFish (changes requested):

ManyTheFish (approved):
Thanks for your contribution!
Bors merge

PR #64, opened 2021-12-17
Count chars instead of graphemes and fix return value when `char_map` is uninitialized

### LINKED ISSUE #63: `num_graphemes_from_bytes` does not work when used for a prefix of a raw Token
# The Issue

The output of `num_graphemes_from_bytes` is wrong when:
 - `num_bytes` is smaller than the length of the string
 - the token does not have the `char_map` initialized - possibly since the Token was created outside of Tokenizer or because the unicode segmenter was not run.

It should return `num_bytes` back since each character is assumed to occupy one byte. Instead, it returns the length of the underlying string.

# Context

This bug was introduced by me in #59 :laughing: 

See also: https://github.com/meilisearch/milli/pull/426#issuecomment-996913963

Samyak2:
~When used with `num_bytes` smaller than the length of the string and when the Token does not have a char_map (created outside of the tokenizer perhaps), the output of `num_graphemes_from_bytes` was wrong - it would return the length of the internal string instead of the number of bytes that was asked.~

Changed `num_graphemes_from_bytes` to `num_chars_from_bytes` because:
- counting graphemes is slower
- graphemes require an external dependency
- only chars are needed for highlighting (which is the intended usage of this function)

When the `char_map` was not initialized (created outside of the tokenizer perhaps), the output of this function was wrong - it would return the length of the internal string instead of the number of bytes that was asked. Fixed this to count the number of chars for given number of bytes, in the string that is available (could be de-unicoded).

# Pull Request

## What does this PR do?
Fixes #63 

## PR checklist
- [x] Does this PR fix an existing issue?

### DIFF
--- a/src/token.rs
+++ b/src/token.rs
@@ -35,7 +35,7 @@ pub struct Token<'a> {
     pub byte_start: usize,
     pub byte_end: usize,
     /// number of bytes used in the normalized string
-    ///  by each grapheme cluster in the original string
+    ///  by each char in the original string
     pub char_map: Option<Vec<usize>>,
 }
 
@@ -73,20 +73,28 @@ impl<'a> Token<'a> {
         self.kind == TokenKind::StopWord
     }
 
-    /// Returns the number of graphemes in original token using number of characters in normalized
+    /// Returns the number of chars in original token using number of bytes in normalized
     /// token.
     ///
-    /// Grapheme clusters are counted in the pre-processed string (just before normalizing).
+    /// chars are counted in the pre-processed string (just before normalizing).
     /// For example, consider the string "Go💼od" which gets normalized to "gobriefcase od".
-    /// `num_graphemes_from_bytes(11)` for this token will return `3` - the number of characters in
+    /// `num_chars_from_bytes(11)` for this token will return `3` - the number of characters in
     /// the original string for 11 bytes in the normalized string.
     ///
     /// # Arguments
     ///
     /// * `num_bytes` - number of bytes in normalized token
-    pub fn num_graphemes_from_bytes(&self, mut num_bytes: usize) -> usize {
+    pub fn num_chars_from_bytes(&self, mut num_bytes: usize) -> usize {
         match &self.char_map {
-            None => num_bytes,
+            None => {
+                // if we don't have a char_map, we look for the number of chars in the current
+                //   (probably normalized) string
+                self.word
+                    .char_indices()
+                    .take_while(|(byte_index, _)| *byte_index <= num_bytes)
+                    .count()
+                    - 1
+            }
             Some(char_map) => char_map
                 .iter()
                 .cloned()
--- a/src/token.rs
+++ b/src/token.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::iter;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum SeparatorKind {
@@ -91,9 +92,27 @@ impl<'a> Token<'a> {
                 //   (probably normalized) string
                 self.word
                     .char_indices()
-                    .take_while(|(byte_index, _)| *byte_index <= num_bytes)
+                    // char_indices returns a tuple of (index, char), we ignore the char
+                    .map(|(char_index, _)| char_index)
+                    // add total number of bytes at the end
+                    .chain(iter::once(self.word.len()))
+                    // need a Vec to use .windows()
+                    .collect::<Vec<usize>>()
+                    // iterate though window of adjacent indices
+                    //   difference of adj indices gives the no. of bytes in that char
+                    //   we added word.len() before to not miss out the last char
+                    .windows(2)
+                    .take_while(|window| {
+                        let char_width = window[1] - window[0];
+                        if num_bytes >= char_width {
+                            num_bytes = num_bytes - char_width;
+                            true
+                        } else {
+                            // stop once we don't have enough bytes to spare
+                            false
+                        }
+                    })
                     .count()
-                    - 1
             }
             Some(char_map) => char_map
                 .iter()
--- a/src/token.rs
+++ b/src/token.rs
@@ -1,5 +1,4 @@
 use std::borrow::Cow;
-use std::iter;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum SeparatorKind {
@@ -82,6 +81,11 @@ impl<'a> Token<'a> {
     /// `num_chars_from_bytes(11)` for this token will return `3` - the number of characters in
     /// the original string for 11 bytes in the normalized string.
     ///
+    /// If the `char_map` hasn't been initialized (it is None), usually done
+    /// by the de-unicoder, it counts the number of characters in self.word
+    /// for the given number of bytes. A char is considered even if the number
+    /// of bytes only covers a portion of it.
+    ///
     /// # Arguments
     ///
     /// * `num_bytes` - number of bytes in normalized token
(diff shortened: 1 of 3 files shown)


ManyTheFish:
I think that your issue is not about returning bytes count instead of char count but about using char instead of graphemes. In [the deunicode normalizer](https://github.com/meilisearch/tokenizer/blob/main/src/normalizer/deunicoder.rs#L31) we may iter over chars `token.word.chars()` instead of  graphemes `token.word.graphemes(true)`. At the end we are interested in how many character we have to highlight. 🤔

ManyTheFish (changes requested):
I think that your issue is not about returning bytes count instead of char count but about using char instead of graphemes. In [the deunicode normalizer](https://github.com/meilisearch/tokenizer/blob/main/src/normalizer/deunicoder.rs#L31) we may iter over chars `token.word.chars()` instead of  graphemes `token.word.graphemes(true)`. In the end we are interested in how many character we have to highlight. 🤔

Samyak2:
> Hey @Samyak2! I think that your issue is not about returning `bytes count` instead of `char count` but about using char instead of graphemes. In [the deunicode normalizer](https://github.com/meilisearch/tokenizer/blob/main/src/normalizer/deunicoder.rs#L31) we may iter over chars `token.word.chars()` instead of graphemes `token.word.graphemes(true)`. In the end we are interested in how many character we have to highlight. thinking

Hmm you're right. If we only need `chars` (and not graphemes), it makes this a lot simpler. Instead of having an extra `char_map`, we can just use [`char_indices`](https://doc.rust-lang.org/std/primitive.str.html#method.char_indices) and `chars()` here.

Thank you! I'll push a commit in this PR fixing these issues.

src/token.rs
```diff
@@ -86,7 +86,7 @@ impl<'a> Token<'a> {
     /// * `num_bytes` - number of bytes in normalized token
     pub fn num_graphemes_from_bytes(&self, mut num_bytes: usize) -> usize {
         match &self.char_map {
-            None => self.word.len(),
+            None => num_bytes,
```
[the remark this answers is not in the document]
Samyak2: `word.len()` always returns byte count. `word.chars().count()` would be needed for char count. So I don't think this is inconsistent.

I will be changing this implementation anyway, so it doesn't matter much :)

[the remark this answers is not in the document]
ManyTheFish: YES! you're right! It should be `self.word.chars().count()` instead of `self.word.len()`, my bad!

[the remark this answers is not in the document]
ManyTheFish: or we rewrite the function into `original_from_normalized_bytes_count` returning only bytes count 🤔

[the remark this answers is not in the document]
Samyak2: I think char count is needed for highlighting - so we should keep it char and not byte count.

Samyak2:
I have made some changes in the latest commit. Instead of graphemes, I have used `.chars()` instead. The `char_map` stores the number of bytes in normalized string for each `char` in original string.

The bug was in the case where `char_map` wasn't initialized. I have modified it to return the number of chars for given bytes in the current string.

src/token.rs
```diff
@@ -73,20 +73,28 @@ impl<'a> Token<'a> {
...
     /// # Arguments
     ///
     /// * `num_bytes` - number of bytes in normalized token
-    pub fn num_graphemes_from_bytes(&self, mut num_bytes: usize) -> usize {
+    pub fn num_chars_from_bytes(&self, mut num_bytes: usize) -> usize {
         match &self.char_map {
-            None => self.word.len(),
+            None => {
+                // if we don't have a char_map, we look for the number of chars in the current
+                //   (probably normalized) string
```
ManyTheFish, line 91: crap! yes! This is always normalized in our case 🤔; but, in another hand, if we don't deunicode, and so, we don't have char_map, there is no issue of counting normalized chars, mmmmmh... 🤔

Let's keep it like this :)

ManyTheFish (approved):

src/token.rs
```diff
@@ -73,20 +73,28 @@ impl<'a> Token<'a> {
...
+                self.word
+                    .char_indices()
+                    .take_while(|(byte_index, _)| *byte_index <= num_bytes)
+                    .count()
+                    - 1
```
ManyTheFish, line 96: why `-1`?
if num_bytes = 1 then count may be 1 and then we return 0

ManyTheFish (changes requested):

src/token.rs
Samyak2: Good catch!

Although, it won't fail when `num_bytes` is 1. The char indices look like this (for "word" string): `[0, 1, 2, 3]`. So the count will be `2` for `num_bytes = 1`.

It will fail for `num_bytes` = `word.len()`. In the above, when `num_bytes` is 4, the function returns 3 which is not right.

src/token.rs
Samyak2: Cool!

src/token.rs
Samyak2: I have fixed this in the latest commit!

ManyTheFish:
Hey @Samyak2 !
I think it would be simplier to do a saturating_sub() on byte_index using each char_index. A take_while() on byte_index > 0 would do the job 🤔

Samyak2:
> Hey @Samyak2 ! I think it would be simplier to do a saturating_sub() on byte_index using each char_index. A take_while() on byte_index > 0 would do the job thinking

What do you mean by `byte_index`?

I initially tried a simple version like this:
```rust
                self.word
                    .char_indices()
                    .take_while(|(char_index, _)| {
                        *char_index < num_bytes
                    })
                    .count()
```

But it does not work for cases where a char is more than one byte, but we don't have enough bytes left to complete the char. For example, `token.num_chars_from_bytes(3)` for the word `Go💼od` should return `2` since we only have enough bytes for 2 characters and the 3rd character will need 2+4 bytes. If we are sure that `num_chars_from_bytes` won't be called in such a case where we only have bytes to represent half (or any fraction) of a character, then we can use this (much) simpler solution.

@ManyTheFish what do you think?

ManyTheFish:
> > Hey @Samyak2 ! I think it would be simplier to do a saturating_sub() on byte_index using each char_index. A take_while() on byte_index > 0 would do the job thinking
> 
> What do you mean by `byte_index`?
> 
> I initially tried a simple version like this:
> 
> ```rust
>                 self.word
>                     .char_indices()
>                     .take_while(|(char_index, _)| {
>                         *char_index < num_bytes
>                     })
>                     .count()
> ```
> 
> But it does not work for cases where a char is more than one byte, but we don't have enough bytes left to complete the char. For example, `token.num_chars_from_bytes(3)` for the word `Go💼od` should return `2` since we only have enough bytes for 2 characters and the 3rd character will need 2+4 bytes. If we are sure that `num_chars_from_bytes` won't be called in such a case where we only have bytes to represent half (or any fraction) of a character, then we can use this (much) simpler solution.
> 
> @ManyTheFish what do you think?

Hello @Samyak2, 

I would say that the below code (your proposal) is acceptable because if I search `gobriefc`   highlighting `Go💼` would be ok.

```rust
                self.word
                   .char_indices()
                   .take_while(|(char_index, _)| {
                       *char_index < num_bytes
                    })
                   .count()
```

But if you think that we should only highlight "overlapped" characters then I would propose.

```rust
                self.word
                   .chars()
                   .take_while(|char| {
                       let stop = char.len() > num_bytes;
                       num_bytes.saturating_sub(char.len());
                       stop
                    })
                   .count()
```

My proposal as an issue, if we have the word `💼` and I'm searching `briefca` nothing would be highlighted 🤔 

what do you think?

Samyak2:
> ```rust
>                 self.word
>                    .chars()
>                    .take_while(|c| {
>                        let stop = c.len() > num_bytes;
>                        num_bytes = num_bytes.saturating_sub(c.len());
>                        stop
>                     })
>                    .count()
> ```
> 
> My proposal as an issue, if we have the word `💼` and I'm searching `briefca` nothing would be highlighted thinking
> 
> what do you think?

@ManyTheFish I had to make some changes to your proposal to make it pass all test cases:
```rust
                self.word
                   .chars()
                   .take_while(|&c| {
                       let stop = num_bytes >= c.len_utf8();
                       num_bytes = num_bytes.saturating_sub(c.len_utf8());
                       stop
                    })
                   .count()
```

But I hadn't thought of this when writing my proposal:

> I would say that the below code (your proposal) is acceptable because if I search gobriefc highlighting Go💼 would be ok.

It makes a lot more sense to highlight it even if only a part of the char is available. I will implement this solution and commit. Thank you!

ManyTheFish (approved):
Tanks @Samyak2!

Bors merge

PR #80, opened 2022-03-14
Update the dependencies

Kerollmops:
This PR upgrades and updates the dependencies of the tokenizer. However, I wasn't able to bump `lindera` and `lindera-core` to _0.11_ as I encountered _once_cell_ poisoned issues.

@ManyTheFish Do you think we can create a new release of the tokenizer, something like a v0.2.8? Bumping dependencies is not breaking, here.


ManyTheFish (approved):
bors merge

PR #83, opened 2022-04-19
Testing and benchmarking

ManyTheFish:
- Add generic tests and benchmarks to ease the implementation of a new segmenter/normalizer and ensure that it's well tested
- Add a test over lifetimes ensuring that the tokenizer is practical to use.


src/segmenter/dummy.example.rs
```diff
@@ -14,7 +14,7 @@ pub struct DummySegmenter;
```
Kerollmops, line 14: Why is the name of the file like so? Why not use underscores here? Like `dummy_example.rs`.

benches/bench.rs
```diff
@@ -1,38 +1,61 @@
...
+            group.bench_with_input(BenchmarkId::new(size, script_lang), &text, |b, text| {
+                b.iter(|| $func(text))
+            });
+        }
+
+        group.finish();
+    };
+}
+
+use benchmark_texts;
```
Kerollmops, line 61: Maybe you can put `pub` in front of the macro?

Kerollmops (changes requested):

Kerollmops (approved):
I love what you have done here!

PR #87, opened 2022-04-27
Implement control character normalizer

ManyTheFish:
Implement control character normalizer removing control characters in the lemma and updating the char map.


Kerollmops (approved):
Thank you very much! LGTM 🎉

PR #92, opened 2022-05-04
Fix API

ManyTheFish:
Change the API to be easier and more natural to use


Kerollmops (approved):
That's awesome, thank you very much!
la bise

bors merge

PR #95, opened 2022-05-05
Improve normalizer dummy

bidoubiwa:
The normalizer dummy had the following issue:

- The normalization of the lemma into lowercase did not work
- The last test was failing
- The documentation was missing a step

Suggestion:

- The normalizer dummy should show an example where a token is separated into multiple tokens. ([rough example here](https://github.com/bidoubiwa/charabia/pull/1))
- Visual on the tests were a bit hard to read. Maybe a pretty error format would be nice.
- Lastly but I think this has already been raised but it could be cool to subscribe to the normalizer from the outside instead of having to add it manually into `mod.rs`


ManyTheFish (approved):
Hey @bidoubiwa, thanks a lot for your feedback.

Bors merge

PR #97, opened 2022-05-05
Make some fixes

ManyTheFish:
#### Fix original lengths method

original_lengths method was returning the char index instead of the char length (char_index + 1) and the byte index was returning the normalized length instead of the original one

#### Refactor script detection

split the text by script in order to use the right specialized pipeline when several scripts are used in the same text like "Ipad 包"


ManyTheFish:
poke @Kerollmops

ManyTheFish:
poke @Kerollmops

src/classifier/classifier.rs
```diff
@@ -53,6 +53,22 @@ where
     }
 }
 
+fn classify_separator(c: char) -> Option<SeparatorKind> {
+    match deunicode_char(c)?.chars().next()? {
+        // Prevent deunicoding cyrillic chars (e.g. ь -> ' is incorrect)
+        _ if ('\u{0410}'..='\u{044f}').contains(&c) => None, // russian cyrillic letters [а-яА-Я]
+        _ if c == '\u{00a0}' => None,                        // non-breaking space
```
Kerollmops, line 60: I would prefer that we consider this non-breaking space a real space for the engine to match more things.
https://en.wikipedia.org/wiki/Non-breaking_space

Kerollmops (changes requested):

Kerollmops (dismissed):

Kerollmops (dismissed):

Kerollmops (approved):
Allez, c'est bon!

ManyTheFish:
bors merge

Issue #99, opened 2022-05-16
Disable char map creation for when we don't need it


Kerollmops:
When we want to index the tokens just to store the words in the inverted index we don't need the char map, maybe we can provide an option to disable creating it.

curquiza:
I put it as good first issue but feel free to ask any questions if you are interested in fixing this. We are aware there is too few information on this issue at the moment, but we are really open to external contributions 😄

matthias-wright:
Is anyone working on this? I would like to give this a try.

matthias-wright:
Hi @ManyTheFish, okay cool! Before I start, I would like to clarify some things.

1. Should I simply add a new method to the Normalizer trait that is called something like `normalize_no_char_map`? I assume that I also should add implementations of this method for LatinNormalizer, ChineseNormalizer, etc.?
2. I will add a new field to the TokenizerBuilder called something like `create_char_map` (bool) and then add a method to set it (like `stop_words`). The default will be true to allow for backwards compatibility. 
3. The same `create_char_map` field is added to Tokenizer. For the `tokenize` method, if the `create_char_map` flag is true, we use `normalize` and if not, we use `normalize_no_char_map`. 

Am I on the right track? :)

ManyTheFish:
Hello @matthias-wright, sorry for the response time.

> In the docstring of Token's [`original_lengths`](https://github.com/meilisearch/charabia/blob/1dcaf3cd872d36ad66be41368ab2a97892980a90/src/token.rs#L128) method, it says that if we call `original_lengths(11)` on the briefcase token in the string "Go💼od", the method will return `(3, 6)`. However, in this snippet:
> 
> ```rust
> use charabia::Tokenize;
> 
> fn main() {
>     let orig = "Go💼od";
>     let mut tokens = orig.tokenize();
>     tokens.next();
>     println!("{:?}", tokens.next().unwrap().original_lengths(11));
> }
> ```
> 
> the output will be `(1, 4)`. Did I confuse something here?

Yes, you are right, the documentation should be updated because the new Latin segmenter behaves differently.
`"Go💼od"`  will now be segmented as  `"go", "💼", "od"`  then normalized  as `"go", "briefcase", "od"`.
For the second token, we return `(1, 4)` meaning 1 character and 4 bytes which corresponds to `💼`

> Hi @ManyTheFish, okay cool! Before I start, I would like to clarify some things.
> 
[...]

PR #109, opened 2022-06-16
Add option to disable char map creation

matthias-wright:
This is a first draft that addresses #99.

To disable the char map creation:
```rust
let s = "Go💼od";
let mut builder = TokenizerBuilder::default();
builder.create_char_map(false);
let tokenizer = builder.build();
let tokens = tokenizer.tokenize(s);
```
The changes are backwards compatible, so the default behavior is to create a char map.

One thing that is awkward right now is that each [`NormalizedTokenIter`](https://github.com/matthias-wright/charabia/blob/0a39d6d6ec374bf5c9ae25b67981aa3f4d405743/src/normalizer/mod.rs#L35) stores a `Option<NormalizerOption>`.
This is caused by the implementation of [`normalize_with_option`](https://github.com/matthias-wright/charabia/blob/0a39d6d6ec374bf5c9ae25b67981aa3f4d405743/src/normalizer/mod.rs#L128), where the normalize method is never explicitly called. Instead the normalize method is called in the `next` method of `NormalizedTokenIter`.
Hence I added a `Option<NormalizerOption>` to `NormalizedTokenIter` and then added a `match` statement to the `next` method that decides to either call `normalize` or `normalize_with_option`.
Of course one improvement would be to use a `&NormalizerOption` in `NormalizedTokenIter` to avoid all the cloning but I am hoping for a better way.

Any feedback and ideas are appreciated :)


matthias-wright:
Hi @curquiza!
This is only a draft because I am not happy that `NormalizedTokenIter` now stores an `Option<NormalizerOption>`.
Changing that will require some more changes and I wanted to discuss possible solutions first before I go ahead, because I am new to the codebase.
Thanks for the heads-up, then I am looking forward to the review in two weeks :smile:

src/normalizer/mod.rs
```diff
@@ -57,13 +65,35 @@ impl<'o> Iterator for NormalizedTokenIter<'o> {
...
     /// Normalize the provided [`Token`].
     ///
     /// A Normalizer can return several `Token`s.
     fn normalize<'o>(&self, token: Token<'o>) -> Box<dyn Iterator<Item = Token<'o>> + 'o>;
 
+    /// Normalize the provided [`Token`].
+    /// Options can be set using the provided [`NormalizerOption`].
+    ///
+    /// A Normalizer can return several `Token`s.
+    fn normalize_with_option<'o>(&self, token: Token<'o>, options: NormalizerOption) -> Box<dyn Iterator<Item = Token<'o>> + 'o>;
```
ManyTheFish, line 95: You can provide a default implementation directly in the trait, that would avoid reimplementing it when it is not necessary.
```suggestion
    fn normalize_with_option<'o>(&self, token: Token<'o>, options: NormalizerOption) -> Box<dyn Iterator<Item = Token<'o>> + 'o> {
        self.normalize(token)
    }
```

ManyTheFish: However, we could always pass the option to the `normalize` method that will avoid having 2 implementations of the same thing.

src/normalizer/mod.rs
```diff
@@ -84,14 +114,33 @@ where
...
+            options: None,
+        };
+        
+        NORMALIZERS.iter().skip(1).fold(first, |token_iter, normalizer| NormalizedTokenIter {
+            token_iter: Box::new(token_iter),
+            inner: Box::new(None.into_iter()),
+            normalizer,
+            options: None,
+        })
+    }
```
ManyTheFish, line 126: we could just call `normalize_with_option` with `NormalizerOption::default()` instead of reimplementing the same thing.
Moreover, this would allow removing the `Option<NormalizerOption>` enum wrapper and directly having a `NormalizerOption`.

src/normalizer/mod.rs
```diff
@@ -36,6 +36,7 @@ pub struct NormalizedTokenIter<'o> {
     token_iter: Box<dyn Iterator<Item = Token<'o>> + 'o>,
     inner: Box<dyn Iterator<Item = Token<'o>> + 'o>,
     normalizer: &'static Box<dyn Normalizer>,
+    options: Option<NormalizerOption>,
```
ManyTheFish, line 39: I think `Option` enum is not necessary when `NormalizerOption` implements `Default`

src/normalizer/mod.rs
```diff
@@ -57,13 +65,35 @@ impl<'o> Iterator for NormalizedTokenIter<'o> {
...
+
+impl NormalizerOption {
+    
+    /// Creates a new [`NormalizerOption`] with default options set.
+    pub fn new() -> NormalizerOption {
+        NormalizerOption {
+            create_char_map: true,
+        }
+    }
+}
```
ManyTheFish, line 82: You should [Implement the `Default` trait](https://doc.rust-lang.org/std/default/trait.Default.html) instead.

src/tokenizer.rs
```diff
@@ -95,11 +95,17 @@ impl<'o> Tokenize<'o, Vec<u8>> for &'o str {
 /// See [`TokenizerBuilder`] to know how to build a [`Tokenizer`].
 pub struct Tokenizer<'sw, A> {
     stop_words: Option<&'sw Set<A>>,
+    create_char_map: bool,
```
ManyTheFish, line 98: You may want to directly wrap it in a `NormalizerOption`

src/tokenizer.rs
```diff
@@ -95,11 +95,17 @@ impl<'o> Tokenize<'o, Vec<u8>> for &'o str {
...
 
 impl<'o, A: AsRef<[u8]>> Tokenizer<'_, A> {
     pub fn tokenize(&self, original: &'o str) -> ClassifiedTokenIter<'o, '_, A> {
-        original.segment().normalize().classify_with_stop_words(self.stop_words)
+        if self.create_char_map {
+            original.segment().normalize().classify_with_stop_words(self.stop_words)
+        } else {
+            let options = NormalizerOption { create_char_map: false };
+            original.segment().normalize_with_option(options).classify_with_stop_words(self.stop_words)
+        }
```
ManyTheFish, line 108: You should always call `normalize_with_option`, the behavior would be the same but the code would be factorized.

src/tokenizer.rs
```diff
@@ -144,14 +150,15 @@ impl<'o, A: AsRef<[u8]>> Tokenizer<'_, A> {
 ///
 pub struct TokenizerBuilder<'sw, A> {
     stop_words: Option<&'sw Set<A>>,
+    create_char_map: bool,
```
ManyTheFish, line 153: You may want to directly wrap it in a `NormalizerOption`

src/tokenizer.rs
```diff
@@ -144,14 +150,15 @@ impl<'o, A: AsRef<[u8]>> Tokenizer<'_, A> {
...
 }
 
 impl<'sw, A> TokenizerBuilder<'sw, A> {
     /// Create a `TokenizerBuilder` with default settings,
     ///
     /// if you don't plan to set stop_words, prefer use [`TokenizerBuilder::default`]
     pub fn new() -> TokenizerBuilder<'sw, A> {
-        Self { stop_words: None }
+        Self { stop_words: None, create_char_map: true }
```
ManyTheFish, line 161: If you directly wrap `create_char_map` it in a `NormalizerOption`, you can do:
```suggestion
        Self { stop_words: None, normalizer_option: NormalizerOption::default() }
```

src/normalizer/latin.rs
```diff
@@ -31,6 +31,24 @@ impl Normalizer for LatinNormalizer {
...
+            let mut lemma = String::new();
+            for c in token.lemma().chars() {
+                // if a char can't be deunicoded, skip it.
+                let deunicoded = deunicode_char(c).unwrap_or("").trim();
+                char_map.as_mut().map(|char_map| char_map.push((c.len_utf8() as u8, deunicoded.len() as u8)));
+                lemma.push_str(&deunicoded);
+            }
+            token.lemma = Cow::Owned(lemma);
+            token.char_map = char_map;
+        }
```
ManyTheFish, line 46: Here, if  `create_char_map` is set to `false` we can directly use [`deunicode`](https://docs.rs/deunicode/latest/deunicode/) on the whole string instead of `deunicode_char` on each character.

ManyTheFish (changes requested):
Despite the factorization, this first implementation seems good and would suit the needs. 👍

matthias-wright:
Thanks for the detailed review!

matthias-wright:
The build failed while compiling `lindera v0.13.5`. I tested it locally and everything worked fine. I am not sure what caused this.

src/normalizer/mod.rs
```diff
@@ -84,13 +75,13 @@ impl NormalizerOption {
 /// Trait defining a normalizer.
 pub trait Normalizer: Sync + Send {
     /// Normalize the provided [`Token`].
+    /// Options can be set using the provided [`NormalizerOption`].
     ///
     /// A Normalizer can return several `Token`s.
-    fn normalize<'o>(&self, token: Token<'o>) -> Box<dyn Iterator<Item = Token<'o>> + 'o>;
+    fn normalize<'o>(&self, token: Token<'o>) -> Box<dyn Iterator<Item = Token<'o>> + 'o> {
+        self.normalize_with_option(token, NormalizerOption::default())
+    }
```
ManyTheFish, line 83: We may remove the `normalize_with_option` method and only keep `normalize` adding a new parameter.
Moreover, I think you can pass the `NormalizerOption` by reference avoiding some clone, but because it only contains a boolean, I'm not sure it's necessary here 🤔 
```suggestion
    fn normalize<'o>(&self, token: Token<'o>, options: NormalizerOption) -> Box<dyn Iterator<Item = Token<'o>> + 'o>;
```

src/normalizer/mod.rs
```diff
@@ -106,38 +97,30 @@ where
...
-            options: None,
-        };
-        
-        NORMALIZERS.iter().skip(1).fold(first, |token_iter, normalizer| NormalizedTokenIter {
-            token_iter: Box::new(token_iter),
-            inner: Box::new(None.into_iter()),
-            normalizer,
-            options: None,
-        })
+        self.normalize_with_option(NormalizerOption::default())
```
ManyTheFish, line 105: :+1:

ManyTheFish (changes requested):

irevoire:
> The build failed while compiling lindera v0.13.5. I tested it locally and everything worked fine. I am not sure what caused this.

Hey @matthias-wright sorry for the issue, if you rebase on main it should work now :+1:

src/normalizer/mod.rs
matthias-wright: Passing `NormalizerOption` by reference caused some issues. For example:
```rust
impl<'o> Tokenize<'o, Vec<u8>> for &'o str {
    fn tokenize(&self) -> ClassifiedTokenIter<'o, '_, Vec<u8>> {
        self.segment().normalize(NormalizerOption::default()).classify()
    }

   ...
}
```
if `normalize()` expects a `&NormalizerOption`, then we can't create a `NormalizerOption` value inside of `tokenize()` and return a reference to it because the reference would be invalid as soon as the call to `tokenize()` ends. We could pass a `&NormalizerOption` to `tokenize()` and simply pass it along to `normalize()` but I think that would break the current API.

For now, I simply derived the `Copy` trait for `NormalizerOption` and then removed all calls to `clone()`. Let me know if that suffices.

src/normalizer/mod.rs
```diff
@@ -71,10 +64,8 @@ pub struct NormalizerOption {
...
 }
 
-impl NormalizerOption {
-    
-    /// Creates a new [`NormalizerOption`] with default options set.
-    pub fn new() -> NormalizerOption {
+impl Default for NormalizerOption {
+    fn default() -> Self {
         NormalizerOption {
             create_char_map: true,
```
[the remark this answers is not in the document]
matthias-wright: Note that I changed the definition of the `test_normalizer!` macro by passing `NormalizerOption { create_char_map: true }` to `normalize()` because several of the tests expect a char map to be created.

src/normalizer/mod.rs
ManyTheFish: While NormalizerOption is a small structure, I'm totally ok with it :+1:

PR #110, opened 2022-07-05
Specify the exact version of lindera we’re using since they broke the compilation on a minor version

irevoire:


ManyTheFish:
Thanks, @irevoire!
bors merge

PR #115, opened 2022-08-18
Optimize thai segmenter

ManyTheFish:
The current Thaï segmenter has a low throughput (< 2MiB/sec) and the dictionary size is over 1MiB.

This PR uses FST encoding to store and search in the dictionary instead of storing the dictionary in raw text and using the segmenter library to search in it.

- The size of the dictionary has been reduced by 41% (from 1,4MiB to 872KiB) 
- The throughput of the dictionary has been increased by 10 (from 1,9MiB/sec to 19MiB/sec)


src/segmenter/thai.rs
```diff
@@ -1,24 +1,74 @@
...
+                    first.len_utf8()
+                }
+            };
+
+            byte_index += length;
+            Some(&tailstr[..length])
+        });
+
+        Box::new(iter)
+    }
```
Kerollmops, line 41:
```suggestion
        let iter = std::iter::from_fn(move || {
            // if we reach the end of the text, we return None.
            if to_segment.is_empty() {
                return None;
            }

            let length = match find_longest_prefix(&WORDS_FST, to_segment.as_bytes()) {
                Some((_, length)) => length,
                None => {
                    // if no sequence matches, we return the next character as a lemma.
                    let first = to_segment.chars().next().unwrap();
                    first.len_utf8()
                }
            };

            let (left, right) = to_segment.split_at(lenght)
            to_segment = right;
            Some(left)
        });

        Box::new(iter)
    }
```

Kerollmops (changes requested):

PR #122, opened 2022-09-06
Add dependabot for GHA

curquiza:
Same as we added in Meilisearch. Only runs once a month. 5 PRs maximum.
https://github.com/meilisearch/meilisearch/blob/main/.github/dependabot.yml


Kerollmops (approved):
bors merge

PR #143, opened 2022-10-04
Implement Pinyin normalizer

### LINKED ISSUE #135: Implement Pinyin normalizer
Today Meilisearch normalizes Chinese characters by converting traditional characters into simplified ones.

#### drawback
This normalization process doesn't seem to enhance the recall of Meilisearch.

#### enhancement
Following [the official discussion about Chinese support](https://github.com/meilisearch/product/discussions/503) in Meilisearch, it is more relevant to normalize Chinese characters by transliterating them into a Phonological version.
In order to have accurate phonology for Mandarin, we should normalize Chinese characters into Pinyin using the [pinyin](https://crates.io/crates/pinyin) crates.

- [/src/normalizer/chinese.rs](https://github.com/meilisearch/charabia/blob/dac6b422e43b666cc232430533e61a50abdd3341/src/normalizer/chinese.rs)

#### Misc
related to [product#503](https://github.com/meilisearch/product/discussions/503)

> Hey! 👋 

crudiedo:
# Pull Request

## Related issue
Fixes #135

## What does this PR do?
This PR replaces `simplified` Chinese normalizer with the `Pinyin` one by using the `pinyin` create.

It also creates `char_map` before converting into `pinyin`, and updates `token.char_end` after doing that since the number of chars will be different. 

It also updates old tests with the new `pinyin` lemmas and `char` lengths.

## PR checklist


src/normalizer/chinese.rs
```diff
@@ -1,26 +1,47 @@
...
+        let mut lemma = String::new();
+
+        // Need to create char_map before converting into Pinyin
+        if options.create_char_map {
+            let mut char_map = Vec::new();
+
+            for c in token.lemma().chars() {
+                let char_len = c.len_utf8() as u8;
+
+                char_map.push((char_len, char_len));
```
ManyTheFish, line 29: the role of the char map is to bind the original char length with its normalized version length,
the char map has to be created at the same time as the normalization.
I suggest merging this for loop with the one below.
Creating a Vec has no cost if you never push in it.
In another hand, you can have an example of how the fill the char map in [the Latin normalizer](https://github.com/meilisearch/charabia/blob/99b713ce0b70e733611d7c1446be48bca7bb19de/src/normalizer/latin.rs#L20-L35).

src/normalizer/chinese.rs
```diff
@@ -1,26 +1,47 @@
...
+            }
+
+            token.char_map = Some(char_map);
         }
 
+        for c in token.lemma().chars() {
+            match c.to_pinyin() {
+                Some(converted) => lemma.push_str(converted.plain()),
```
ManyTheFish, line 37:
```suggestion
                Some(converted) => lemma.push_str(converted.with_tone()),
```
I'd prefer to keep the tone information at the step 😄

Cargo.toml
```diff
@@ -22,6 +22,7 @@ slice-group-by = "0.3.0"
 unicode-segmentation = "1.6.0"
 whatlang = "0.16.1"
 lindera = { version = "=0.16.0", features = ["ipadic"], optional = true }
+pinyin = "0.9"
```
ManyTheFish, line 25: Could you please:
1) keep only the features we need
```suggestion
pinyin = { version = "0.9", default-features = false, features = ["with_tone"], optional = true }
```
2) remove the upper `character_converter` dependency, it is no more needed 😄
3) replace `"dep:character_converter"` by `"dep:pinyin"` in the below `chinese` feature

ManyTheFish (changes requested):
Hello @crudiedo!
Thank you for your contribution, I requested some changes in your PR,
I'll re-review your PR when you have fixed them! 👍

src/normalizer/chinese.rs
crudiedo: Thank you for your review - the char map role was a bit unclear for me, but after your explanation I finally got it (:

All the changes should be addressed now.

src/normalizer/chinese.rs
```diff
@@ -1,26 +1,51 @@
...
+                        .map(|char_map| char_map.push((c.len_utf8() as u8, c.len_utf8() as u8)));
+
+                    lemma.push(c);
+                }
+            }
         }
 
+        token.lemma = Cow::Owned(lemma);
+        token.char_end = token.lemma.chars().count();
+        token.byte_end = token.lemma.len();
```
ManyTheFish, line 46: These values are kept in the token in the case the token is normalized, if we want to know the normalized lengths, we can dynamically compute it at any time.
```suggestion
```

ManyTheFish (changes requested):
I made a last change request.

Then it will be good for me! 😄

ManyTheFish (approved):
Perfect!
I let `bors` run the tests, and if everything is good, your PR will be automatically merged.
Thank you for your contribution!

bors merge

PR #147, opened 2022-10-07
Separate out FstSegmenter from ThaiSegmenter

### LINKED ISSUE #130: Move the FST based Segmenter in a standalone file
For the Thaï segmenter, we tried a Final-state-transducer (FST) based segmenter.
This segmenter has really good performance and the dictionaries encoded as FSTs are smaller than raw `txt`/`csv`/`tsv` dictionaries.
For now, the segmenter is in the Thaï segmenter file ([segmenter/thai.rs](https://github.com/meilisearch/charabia/blob/927b0d171286e05aaa938f0dd43ec9b84470a05e/src/segmenter/thai.rs#L42-L70)), and, in order to reuse it for other Languages, it would be better to move this segmenter to its own file.
A new struct `FstSegmenter` may be created wrapping all the iterative segmentation logic.

### File expected to be modified
- [segmenter/thai.rs](https://github.com/meilisearch/charabia/blob/main/src/segmenter/thai.rs)
- create [segmenter/utils.rs](https://github.com/meilisearch/charabia/blob/main/src/segmenter/)

> Hey! 👋 

daniel-shuy:
# Pull Request

## Related issue
Closes #130

## What does this PR do?
- Separate out `FstSegmenter` from `ThaiSegmenter` into its own file (`/segmenter/utils.rs`)

## PR checklist

### DIFF
--- a/src/segmenter/utils.rs
+++ b/src/segmenter/utils.rs
@@ -1,18 +1,18 @@
 use fst::raw::{Fst, Output};
 
 /// Final-state-transducer (FST) Segmenter
-pub(crate) struct FstSegmenter<'o> {
-    words_fst: &'o Fst<&'o [u8]>
+pub(crate) struct FstSegmenter<'fst> {
+    words_fst: &'fst Fst<&'fst [u8]>
 }
 
-impl<'o> FstSegmenter<'o> {
-    pub(crate) fn new(words_fst: &'o Fst<&'o [u8]>) -> Self {
+impl<'fst> FstSegmenter<'fst> {
+    pub(crate) fn new(words_fst: &'fst Fst<&'fst [u8]>) -> Self {
         Self { words_fst }
     }
 }
 
-impl<'o> FstSegmenter<'o> {
-    pub fn segment_str(&'o self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
+impl<'fst> FstSegmenter<'fst> {
+    pub fn segment_str(&'fst self, mut to_segment: &'fst str) -> Box<dyn Iterator<Item = &'fst str> + 'fst> {
         let iter = std::iter::from_fn(move || {
             // if we reach the end of the text, we return None.
             if to_segment.is_empty() {
--- a/src/segmenter/utils.rs
+++ b/src/segmenter/utils.rs
@@ -9,9 +9,7 @@ impl<'fst> FstSegmenter<'fst> {
     pub(crate) fn new(words_fst: &'fst Fst<&'fst [u8]>) -> Self {
         Self { words_fst }
     }
-}
 
-impl<'fst> FstSegmenter<'fst> {
     pub fn segment_str<'o>(
         &'fst self,
         mut to_segment: &'o str
--- a/src/segmenter/utils.rs
+++ b/src/segmenter/utils.rs
@@ -12,7 +12,12 @@ impl<'fst> FstSegmenter<'fst> {
 }
 
 impl<'fst> FstSegmenter<'fst> {
-    pub fn segment_str(&'fst self, mut to_segment: &'fst str) -> Box<dyn Iterator<Item = &'fst str> + 'fst> {
+    pub fn segment_str<'o>(
+        &'fst self,
+        mut to_segment: &'o str
+    ) -> Box<dyn Iterator<Item = &'o str> + 'o>
+        where
+            'fst: 'o {
         let iter = std::iter::from_fn(move || {
             // if we reach the end of the text, we return None.
             if to_segment.is_empty() {
--- a/src/segmenter/utils.rs
+++ b/src/segmenter/utils.rs
@@ -1,0 +1,68 @@
+use fst::raw::{Fst, Output};
+
+/// Final-state-transducer (FST) Segmenter
+pub(crate) struct FstSegmenter<'o> {
+    words_fst: &'o Fst<&'o [u8]>
+}
+
+impl<'o> FstSegmenter<'o> {
+    pub(crate) fn new(words_fst: &'o Fst<&'o [u8]>) -> Self {
+        Self { words_fst }
+    }
+}
+
+impl<'o> FstSegmenter<'o> {
+    pub fn segment_str(&'o self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
+        let iter = std::iter::from_fn(move || {
+            // if we reach the end of the text, we return None.
+            if to_segment.is_empty() {
+                return None;
+            }
+
+            let length = match find_longest_prefix(self.words_fst, to_segment.as_bytes()) {
+                Some((_, length)) => length,
+                None => {
+                    // if no sequence matches, we return the next character as a lemma.
+                    let first = to_segment.chars().next().unwrap();
+                    first.len_utf8()
+                }
+            };
+
+            let (left, right) = to_segment.split_at(length);
+            to_segment = right;
+            Some(left)
+        });
+
+        Box::new(iter)
+    }
+}
+
+/// Thanks to @llogiq for this function
+/// https://github.com/BurntSushi/fst/pull/104/files
+///
+/// find the longest key that is prefix of the given value.
+///
+/// If the key exists, then `Some((value, key_len))` is returned, where
+/// `value` is the value associated with the key, and `key_len` is the
+/// length of the found key. Otherwise `None` is returned.
+///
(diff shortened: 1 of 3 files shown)


src/segmenter/utils.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+impl<'o> FstSegmenter<'o> {
+    pub(crate) fn new(words_fst: &'o Fst<&'o [u8]>) -> Self {
+        Self { words_fst }
+    }
+}
+
+impl<'o> FstSegmenter<'o> {
+    pub fn segment_str(&'o self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
```
daniel-shuy, line 15: I would've like `FstSegmenter` to implement the `Segmenter` trait, e.g.
```rs
impl<'o> Segmenter for FstSegmenter<'o> {
```
but because the `FstSegmenter` struct has its own lifetime (to keep the FST alive for the lifetime of the struct), and the `Segmenter::segment_str` function signature also has its own lifetime:
https://github.com/meilisearch/charabia/blob/2beae7e18293f3de023181cd9dac4aab759b7a34/src/segmenter/mod.rs#L145-L148
I'm not sure how to do so without modifying the `Segmenter` trait (I'm relatively new to Rust).

If I use an anonymous lifetime, e.g.
```rs
impl Segmenter for FstSegmenter<'_> {
    fn segment_str<'o>(&self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
```
it fails with:
```
error: lifetime may not live long enough
  --> src/segmenter/utils.rs:38:9
   |
17 |     fn segment_str<'o>(&self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
   |                    --  ----- has type `&FstSegmenter<'1>`
   |                    |
   |                    lifetime `'o` defined here
...
38 |         Box::new(iter)
   |         ^^^^^^^^^^^^^^ associated function was supposed to return data with lifetime `'o` but it is returning data with lifetime `'1`
```

ManyTheFish: Good question,
first, in this codebase, the lifetime `'o` is related to the **original** text, for example, the [`Token` has the lifetime `'o`](https://github.com/meilisearch/charabia/blob/6d0a01d371088283d4e9bbadba24c0a0d2344571/src/token.rs#L35-L53) because it keeps a reference of the **original** text via the field `lemma`.

In your case, the struct `FstSegmenter` doesn't store any references to the **original** text, so you shouldn't name your lifetime `'o` but `'d` or `'wd` for **word dictionary** or even `'fst`.

But, does this will fix your issue? I'm not completely sure, because [from_fn](https://doc.rust-lang.org/std/iter/fn.from_fn.html) struggles to figure out which lifetime is returned, and you would probably have to precise it:
```rust
let iter = std::iter::from_fn::<&'o str, _>(move || {
```
But I'm not 100% sure that would fix your issue 🤔

ManyTheFish (commented):
Hey @daniel-shuy, I responded to your comment suggesting the modification to continue.
However, your current work is valid if you change the name of the lifetime you used.
It is up to you to continue the implementation 😄

src/segmenter/utils.rs
daniel-shuy: Ah, I wasn't sure what the `'o` meant, thanks for clarifying it, I've renamed it to `'fst` as you suggested.

daniel-shuy: The lifetime name is not the cause of the issue though, even after renaming it, it still fails with the same `lifetime may not live long enough` error.

```rs
impl<'fst> Segmenter for FstSegmenter<'fst> {
    fn segment_str<'o>(&self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
        let iter = std::iter::from_fn::<&'o str, _>(move || {
```
fails with the same error as well. `<&'o str, _>` is exactly what is inferred. The issue is not that `from_fn` cannot figure out which lifetime is returned, but that `'o` might outlive `'fst`, since `'o` is determined by the caller.

We also cannot declare `'o` as a subtype of `fst` (to ensure that `'fst` outlives `'o`), e.g.
```rs
impl<'fst> Segmenter for FstSegmenter<'fst> {
    fn segment_str<'o: 'fst>(&self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
```
because if the caller uses the `Segmenter` trait, then there is no guarantee that the `'o` lifetime will outlive `'fst`, since the `Segmenter` trait doesn't have the `'fst` lifetime.

So it seems that the only way for `FstSegmenter` to implement `Segmenter` is to modify the `Segmenter` trait, by moving the lifetime from the function to the trait, i.e.
```rs
pub trait Segmenter<'o>: Sync + Send {
    fn segment_str(&self, s: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o>;
}
```
Unfortunately, that would require changing all implementations of `Segmenter`, therefore I think its better for `FstSegmenter` to not implement `Segmenter`.

src/segmenter/utils.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+impl<'fst> FstSegmenter<'fst> {
+    pub(crate) fn new(words_fst: &'fst Fst<&'fst [u8]>) -> Self {
+        Self { words_fst }
+    }
+}
+
+impl<'fst> FstSegmenter<'fst> {
+    pub fn segment_str(&'fst self, mut to_segment: &'fst str) -> Box<dyn Iterator<Item = &'fst str> + 'fst> {
```
ManyTheFish, line 15: Well, there is an issue here, `self` has the lifetime `'fst` which is good because it stores internally a reference to the dictionary. However `to_segment` is not related to the dictionary and so shouldn't have the `'fst` lifetime. But `to_segmentù needs a lifetime, this lifetime is the one related to the **original** text: `'o`, and because it is not needed for `FstSegmenter` but is needed for `segment_str` we will pass it locally:
```suggestion
    pub fn segment_str<'o>(&'fst self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
```

ManyTheFish (changes requested):

src/segmenter/utils.rs
daniel-shuy: If we do that then we run into the same issue again, `'fst` may not outlive `'o`.

If I am not mistaken, using `'fst` for `to_segment` does not mean that it will have the same lifetime as the FST, but that the FST must have a longer lifetime than `to_segment`.

Based on my understanding of lifetimes, it is equivalent to:
```rs
pub fn segment_str<'o: 'fst>(&'fst self, mut to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'fst> {
```
where `'o: 'fst` constraints `'o` to be shorter than `'fst`

ManyTheFish: Yes you're right, but, I'd prefer to have an explicit `where 'fst: 'o` at the end of `segment_str` instead of putting the same name for every lifetime 😬

daniel-shuy: Done! I learnt a lot about Rust from this, thanks!

src/segmenter/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+    words_fst: &'fst Fst<&'fst [u8]>
+}
+
+impl<'fst> FstSegmenter<'fst> {
+    pub(crate) fn new(words_fst: &'fst Fst<&'fst [u8]>) -> Self {
+        Self { words_fst }
+    }
+}
+
+impl<'fst> FstSegmenter<'fst> {
```
ManyTheFish, line 14: A small last change, let's merge these two `impl` 
```suggestion
impl<'fst> FstSegmenter<'fst> {
    pub(crate) fn new(words_fst: &'fst Fst<&'fst [u8]>) -> Self {
        Self { words_fst }
    }

```

ManyTheFish (changes requested):
I requested a last small change,
after that, your PR will be merged! 🎉

src/segmenter/utils.rs
daniel-shuy: Oh right, forgot there's no need to have multiple impl blocks now that there's no anonymous lifetime, I've unified the impl blocks, thanks!

ManyTheFish (approved):
Hello @daniel-shuy,
it looks good to me!
I let bors run the tests, and if everything passes, your PR will be merged automatically.
Thank you for your contribution! 🙏 

bors merge

PR #149, opened 2022-10-08
Add Japanese normalizer to cover Katakana to Hiragana

### LINKED ISSUE #131: Implement a Japanese specialized Normalizer
Today, there is no specialized normalizer for the Japanese Language.

### drawback
Meilisearch is unable to find the hiragana version of a word with a katakana query, for instance, `ダメ`, is also spelled `駄目`, or `だめ`

### Technical approach
Create a new Japanese normalizer that unifies hiragana and katakana equivalences.

#### Interesting libraries
- [wana_kana](https://crates.io/crates/wana_kana) seems promising to convert everything in Hiragana

- create  [/src/normalizer/japanese.rs](https://github.com/meilisearch/charabia/blob/dac6b422e43b666cc232430533e61a50abdd3341/src/normalizer/)
- [/src/normalizer/mod.rs](https://github.com/meilisearch/charabia/blob/dac6b422e43b666cc232430533e61a50abdd3341/src/normalizer/mod.rs)

### Misc
related to [product#532](https://github.com/meilisearch/product/discussions/532)

> Hey! 👋 

choznerol:
# Pull Request

## Related issue
Fixes #131

## What does this PR do?
- Add a new Normalizer for Japanese, which converts Katakana to Hiragana

<div id=limitations ></div>

### Converting from Kanji is not supported

From #131:
> ... for instance, `ダメ`, is also spelled `駄目`, or `だめ`
> ... [wana_kana](https://crates.io/crates/wana_kana) seems promising to convert everything in Hiragana

After some experiments and checking [convert options](https://docs.rs/wana_kana/2.1.0/wana_kana/struct.Options.html), it seems like [wana_kana](https://crates.io/crates/wana_kana) does not support converting Kanji to Hiragana or Romaji. For example:
- `to_hiragana("ダメ駄目だめ")` will be `"だめ駄目だめ"`
- `to_romaji("ダメ駄目だめ")` will be `"dame駄目dame"`

## PR checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ unicode-segmentation = "1.6.0"
 whatlang = "0.16.1"
 lindera = { version = "=0.16.0", features = ["ipadic"], optional = true }
 pinyin = { version = "0.9", default-features = false, features = ["with_tone"], optional = true }
+wana_kana = "2.1.0"
 
 [features]
 default = ["chinese", "hebrew", "japanese", "thai"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,10 @@ slice-group-by = "0.3.0"
 unicode-segmentation = "1.6.0"
 whatlang = "0.16.1"
 lindera = { version = "=0.16.0", features = ["ipadic"], optional = true }
-pinyin = { version = "0.9", default-features = false, features = ["with_tone"], optional = true }
-wana_kana = "2.1.0"
+pinyin = { version = "0.9", default-features = false, features = [
+  "with_tone",
+], optional = true }
+wana_kana = { version = "2.1.0", optional = true }
 
 [features]
 default = ["chinese", "hebrew", "japanese", "thai"]
@@ -35,6 +37,7 @@ hebrew = []
 
 # allow japanese specialized tokenization
 japanese = ["dep:lindera"]
+japanese-transliteration = ["dep:wana_kana"]
 
 # allow thai specialized tokenization
 thai = []
--- a/src/normalizer/japanese.rs
+++ b/src/normalizer/japanese.rs
@@ -1,0 +1,194 @@
+use std::borrow::Cow;
+
+use super::{Normalizer, NormalizerOption};
+use crate::detection::{Language, Script};
+use crate::Token;
+use wana_kana::is_hiragana::*;
+use wana_kana::to_hiragana::to_hiragana_with_opt;
+use wana_kana::Options;
+
+/// Japanese specialized [`Normalizer`].
+///
+/// This Normalizer uses [`wana_kana`] internally to convert Katakana and Kanji to Hiragana
+pub struct JapaneseNormalizer;
+
+impl Normalizer for JapaneseNormalizer {
+    fn normalize<'o>(
+        &self,
+        mut token: Token<'o>,
+        options: NormalizerOption,
+    ) -> Box<dyn Iterator<Item = Token<'o>> + 'o> {
+        if is_hiragana(token.lemma()) {
+            // No need to convert
+
+            if options.create_char_map && token.char_map.is_none() {
+                let mut char_map = Vec::new();
+                for c in token.lemma().chars() {
+                    char_map.push((c.len_utf8() as u8, c.len_utf8() as u8));
+                }
+                token.char_map = Some(char_map);
+            }
+        } else {
+            // Convert Katakana to Hiragana
+            let new_lemma = to_hiragana_with_opt(
+                token.lemma(),
+                Options {
+                    pass_romaji: true, // Otherwise 'ダメ駄目だめHi' would become 'だめ駄目だめひ'
+                    ..Default::default()
+                },
+            );
+
+            if options.create_char_map && token.char_map.is_none() {
+                debug_assert!(
+                    token.lemma().len() == new_lemma.len(),
+                    concat!(
+                        r#"`to_hiragana` changed the lemma len from {} to {} but the current `char_map` computation "#,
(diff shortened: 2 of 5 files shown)


src/normalizer/japanese.rs
```diff
@@ -0,0 +1,194 @@
...
+                    token.lemma().len() == new_lemma.len(),
+                    concat!(
+                        r#"`to_hiragana` changed the lemma len from {} to {} but the current `char_map` computation "#,
+                        r#"expected them to be equal. If `to_hiragana` does change len of char somehow, consider "#,
+                        r#"calling `to_hiragana(char)` char by char instead of only calling `to_hiragana(lemma)` once."#
+                    ),
+                    token.lemma().len(),
+                    new_lemma.len()
+                );
+                let old_new_chars = token.lemma().chars().zip(new_lemma.chars());
```
choznerol, line 52: Unlike `to_pinyin` operates on each `char`, `to_hiragana()` can operate on the whole `lemma` at once. This provides some performance benefit because each `to_hiragana()` call has some [branching cost](https://docs.rs/wana_kana/2.1.0/src/wana_kana/to_hiragana.rs.html#24-32). However, `to_hiragana(lemma)` instead of `to_hiragana(char)` also makes the `char_map` implementation a bit more complicated, and I am not sure about the trade off here.

src/normalizer/japanese.rs
```diff
@@ -0,0 +1,194 @@
...
+            // No need to convert
+
+            if options.create_char_map && token.char_map.is_none() {
+                let mut char_map = Vec::new();
+                for c in token.lemma().chars() {
+                    char_map.push((c.len_utf8() as u8, c.len_utf8() as u8));
+                }
+                token.char_map = Some(char_map);
+            }
+        } else {
```
ManyTheFish, line 31: Having an `Identity` char map is the same that not having it, so don't bother creating it.
```suggestion
        if !is_hiragana(token.lemma()) {
```

ManyTheFish (changes requested):
Hello @choznerol,
I requested some changes to your PR, this would simplify the implementation.
However, the general PR is great, thanks!

src/normalizer/japanese.rs
choznerol: Oh, I originally thought we must create one when `create_char_map` is true 😅  Seems this is not the case, so the implementation can be much simpler. Thanks for the suggestion!

choznerol:
> you seem to be 100% sure that this conversion will always keep the same length as the original character.

I only manually tested the mappings I found in the crate repo. To be more confident about the assertion, I searched the unit tests about `to_hiragana` and assert their `len()` unchanged, and all the tests still pass: https://github.com/PSeitz/wana_kana_rust/commit/805c9dee288989fb1eb5ce8083c79102f222ef89

ManyTheFish:
@choznerol, in this case just ignore the char_map creation for the implementation of this Normalizer 👍

src/normalizer/japanese.rs
choznerol: Addressed in 49bc172

Cargo.toml
```diff
@@ -22,6 +22,7 @@ unicode-segmentation = "1.6.0"
 whatlang = "0.16.1"
 lindera = { version = "=0.16.0", features = ["ipadic"], optional = true }
 pinyin = { version = "0.9", default-features = false, features = ["with_tone"], optional = true }
+wana_kana = "2.1.0"
```
ManyTheFish, line 25:
```suggestion
wana_kana = { version = "2.1.0", optional = true}
```

Cargo.toml
```diff
@@ -22,6 +22,7 @@ unicode-segmentation = "1.6.0"
...
 
 [features]
 default = ["chinese", "hebrew", "japanese", "thai"]
```
ManyTheFish, line 28:
```suggestion
default = ["chinese", "hebrew", "japanese", "thai"]

# allow japanese character transliteration (put this under line 38)
japanese-transliteration  = ["dep:wana_kana"]
```

src/normalizer/mod.rs
```diff
@@ -3,6 +3,8 @@ use once_cell::sync::Lazy;
 #[cfg(feature = "chinese")]
 pub use self::chinese::ChineseNormalizer;
 pub use self::control_char::ControlCharNormalizer;
+#[cfg(feature = "japanese")]
```
ManyTheFish, line 6:
```suggestion
#[cfg(feature = "japanese-transliteration")]
```

src/normalizer/mod.rs
```diff
@@ -12,6 +14,8 @@ use crate::Token;
 #[cfg(feature = "chinese")]
 mod chinese;
 mod control_char;
+#[cfg(feature = "japanese")]
```
ManyTheFish, line 17:
```suggestion
#[cfg(feature = "japanese-transliteration")]
```

src/normalizer/mod.rs
```diff
@@ -22,6 +26,8 @@ pub static NORMALIZERS: Lazy<Vec<Box<dyn Normalizer>>> = Lazy::new(|| {
         Box::new(LowercaseNormalizer),
         #[cfg(feature = "chinese")]
         Box::new(ChineseNormalizer),
+        #[cfg(feature = "japanese")]
```
ManyTheFish, line 29:
```suggestion
        #[cfg(feature = "japanese-transliteration")]
```

ManyTheFish (changes requested):
Requesting changes [related to the discussion on the dedicated issue](https://github.com/meilisearch/charabia/issues/131#issuecomment-1280902225).

Thanks a lot for your investment in this work! 👍

PR #152, opened 2022-10-11
impl name and from_name on enums

ManyTheFish:


Kerollmops (approved):
Not sure if these two `unwrap_or_default` are secure as they could hide a bug but if you tell me the contrary, let's merge!

Kerollmops (approved):
I like what you've done here!
bors merge

PR #155, opened 2022-10-22
Test thai homographs

Roms1383:
# Pull Request

This simplistic PR add thai-based homographs (a.k.a คําพ้องรูป) to tests.

## Related issue
No related issue

## What does this PR do?
- change `test_segmenter!` to allow for multiple sets of tests
- add 2 thai sentences made of homographs
   > sentences were co-authored by thai native (@LittleKitty2832)

## PR checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,11 +21,8 @@ slice-group-by = "0.3.0"
 unicode-segmentation = "1.6.0"
 whatlang = "0.16.1"
 lindera = { version = "=0.17.0", default-features = false, optional = true }
-pinyin = { version = "0.9", default-features = false, features = [
-  "with_tone",
-], optional = true }
+pinyin = { version = "0.9", default-features = false, features = ["with_tone"], optional = true }
 wana_kana = { version = "2.1.0", optional = true }
-paste = "1.0.9"
 
 [features]
 default = ["chinese", "hebrew", "japanese", "thai", "korean"]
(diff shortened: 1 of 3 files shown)


Roms1383:
As a side note, we initially thought "สระผม" should be 2 different words ("สระ" `to wash` and "ผม" `hair` in this context), but apparently [dictionary also states that it's a single word](http://www.thai-language.com/id/202209). So we updated the test accordingly and it passes.

Roms1383:
> I'm wondering why it is needed to change the test macro and not simply write 2 sentences in the same string. 🤔
shouldn't it be simpler?

Sure @ManyTheFish I can revert the changes in the macros and concatenate the sentences altogether if you want.

> About the dictionary, I agree with the fact that a lot of words could be split into smaller parts, however, as a non Thaï speaker, it's difficult for me to clean the dictionary without making errors.

Well, I've checked again on [thai-language](http://www.thai-language.com/) and [longdo](https://dict.longdo.com/) dictionaries they both give the same results as the tokenizer, so I think everything's fine indeed.

If sometimes you need some help for translation, we'll be happy to assist with @LittleKitty2832.

Roms1383:
ok @ManyTheFish, changes done :)

ManyTheFish (approved):
Perfect! let's merge this!

About dictionaries, the issue in Information Retrieval is that we prefer having searchable data than having accurate data, for instance, `สำนักงานคณะกรรมการป้องกันและปราบปรามการทุจริตและประพฤติมิชอบในวงราชการ` is considered as only one word. It's probably accurate but if I search for `ปราบปรามการทุจริต`, should I find `สำนักงานคณะกรรมการป้องกันและปราบปรามการทุจริตและประพฤติมิชอบในวงราชการ`? If the answer is yes, we should think of sub-splitting this word in order to be "more searchable".

Anyway! Thank you for your contribution, I let bors run the test and if everything goes well, your PR will be automatically merged.

bors merge

PR #157, opened 2022-10-24
Simplify normalizer implementation

### LINKED ISSUE #156: Refactor normalizers
Today creating a normalizer is way harder than creating a segmenter, this is mainly due to the **char map**, a necessary field to manage highlights.

## Technical Approach

Refactor the `Normalizer` trait to implement a `normalize_str` and `normalize_char` method that takes a `Cow<str>` in parameter and return a `Cow<str>`. All the char map creation should be done in a function calling these two methods.

ManyTheFish:
- Create a trait `CharNormalizer`:
  - Add a method `normalize_char` to the `CharNormalizer` trait that needs to be implemented in each normalizer
  - Add a method `normalize_str` to the `CharNormalizer` trait with a default implementation that normalizes a provided string without bothering with the `char_map` calling the method `normalize_char`.

- the method `normalize` of  the `Normalizer` trait now has a default implementation for the `CharNormalizer` trait that handles the `char_map` creation and calls `normalize_str` to normalize the token
  - this avoids implementing the char_map computing for each normalizer which will ease the future contributions 
  - this can be overridden if the char_map computing is useless for a normalizer
- the method `should_normalize` takes now a reference to the whole `Token` in parameters extending the normalizing conditions. 

Fix #156

### DIFF
--- a/src/normalizer/lowercase.rs
+++ b/src/normalizer/lowercase.rs
@@ -1,7 +1,5 @@
 use std::borrow::Cow;
 
-use cow_utils::CowUtils;
-
 use super::{Normalizer, NormalizerOption};
 use crate::detection::{Language, Script};
 use crate::Token;
@@ -11,18 +9,22 @@ use crate::Token;
 pub struct LowercaseNormalizer;
 
 impl Normalizer for LowercaseNormalizer {
-    fn normalize<'o>(
-        &self,
-        mut token: Token<'o>,
-        _options: NormalizerOption,
-    ) -> Box<dyn Iterator<Item = Token<'o>> + 'o> {
-        // Cow::Borrowed holds a reference to token, which makes it impossible to directly replace
-        // word with the `cow_to_lowercase` result
-        if let Cow::Owned(s) = token.lemma.cow_to_lowercase() {
-            token.lemma = Cow::Owned(s);
+    // lowercasing characters doesn't change the characters length,
+    // so the `normalize` method is overloaded to skip the useless char_map computing.
+    fn normalize<'o>(&self, mut token: Token<'o>, _options: NormalizerOption) -> Token<'o> {
+        if let Cow::Owned(lemma) = self.normalize_str(token.lemma()) {
+            token.lemma = Cow::Owned(lemma);
         }
 
-        Box::new(Some(token).into_iter())
+        token
+    }
+
+    fn normalize_str<'o>(&self, src: &'o str) -> Cow<'o, str> {
+        if src.chars().any(char::is_uppercase) {
+            Cow::Owned(src.to_lowercase())
+        } else {
+            Cow::Borrowed(src)
+        }
     }
 
     fn should_normalize(&self, script: Script, _language: Option<Language>) -> bool {
--- a/src/normalizer/lowercase.rs
+++ b/src/normalizer/lowercase.rs
@@ -12,15 +12,9 @@ impl Normalizer for LowercaseNormalizer {
     // lowercasing characters doesn't change the characters length,
     // so the `normalize` method is overloaded to skip the useless char_map computing.
     fn normalize<'o>(&self, mut token: Token<'o>, _options: NormalizerOption) -> Token<'o> {
-        if let Cow::Owned(lemma) = self.normalize_str(token.lemma()) {
-            token.lemma = Cow::Owned(lemma);
-        }
+        token.lemma = Cow::Owned(token.lemma().to_lowercase());
 
         token
-    }
-
-    fn normalize_str<'o>(&self, src: &'o str) -> Cow<'o, str> {
-        Cow::Owned(src.to_lowercase())
     }
 
     fn should_normalize(&self, token: &Token) -> bool {
--- a/src/normalizer/lowercase.rs
+++ b/src/normalizer/lowercase.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 
 use super::{Normalizer, NormalizerOption};
-use crate::detection::{Language, Script};
+use crate::detection::Script;
 use crate::Token;
 
 /// A global [`Normalizer`] lowercasing characters.
@@ -20,16 +20,13 @@ impl Normalizer for LowercaseNormalizer {
     }
 
     fn normalize_str<'o>(&self, src: &'o str) -> Cow<'o, str> {
-        if src.chars().any(char::is_uppercase) {
-            Cow::Owned(src.to_lowercase())
-        } else {
-            Cow::Borrowed(src)
-        }
+        Cow::Owned(src.to_lowercase())
     }
 
-    fn should_normalize(&self, script: Script, _language: Option<Language>) -> bool {
+    fn should_normalize(&self, token: &Token) -> bool {
         // https://en.wikipedia.org/wiki/Letter_case#Capitalisation
-        matches!(script, Script::Latin | Script::Cyrillic | Script::Greek | Script::Georgian)
+        matches!(token.script, Script::Latin | Script::Cyrillic | Script::Greek | Script::Georgian)
+            && token.lemma.chars().any(char::is_uppercase)
     }
 }
 
--- a/src/normalizer/mod.rs
+++ b/src/normalizer/mod.rs
@@ -9,7 +9,6 @@ pub use self::control_char::ControlCharNormalizer;
 pub use self::japanese::JapaneseNormalizer;
 pub use self::latin::LatinNormalizer;
 pub use self::lowercase::LowercaseNormalizer;
-use crate::detection::{Language, Script};
 use crate::normalizer::nonspacing_mark::NonspacingMarkNormalizer;
 use crate::Token;
 
@@ -48,7 +47,7 @@ impl<'o> Iterator for NormalizedTokenIter<'o> {
 
     fn next(&mut self) -> Option<Self::Item> {
         let token = self.token_iter.next()?;
-        if self.normalizer.should_normalize(token.script, token.language) {
+        if self.normalizer.should_normalize(&token) {
             Some(self.normalizer.normalize(token, self.options))
         } else {
             Some(token)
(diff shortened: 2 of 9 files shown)


src/normalizer/mod.rs
```diff
@@ -78,17 +72,51 @@ pub trait Normalizer: Sync + Send {
...
+                Some(mut char_map) => {
+                    let mut lemma = String::new();
+                    let mut tail = token.lemma.as_ref();
+                    for (_, normalized_len) in char_map.iter_mut() {
+                        let (head, t) = tail.split_at(*normalized_len as usize);
+                        tail = t;
+                        let normalized = self.normalize_str(head);
+                        *normalized_len = normalized.len() as u8;
+                        lemma.push_str(normalized.as_ref());
+                    }
```
loiclec, line 87: Trying to understand what is happening here, for myself:
1. We look at every character of the original lemma.
2. For every character, we normalise it using `normalise_str`
3. Then we change the length of this character in the `char_map`, based on the length of the normalised char
4. And we push the normalised char to the new lemma

That looks good :-)

I wonder, however, what a "character" from the `char_map` is. Is it always a single unicode code point (a `char` in Rust) or can it be a grapheme cluster? 

In either case, I think there is a risk of misunderstanding about the kind of inputs that `normalise_str` accepts. I initially expected it to accept whole strings such as `lion`, but it seems like it is mostly given single characters unless `options.create_char_map` is false.

If a character in the `char_map` is guaranteed to be a `char`, then maybe it would be worth having a `normalize_char` trait function? It could clarify that a normaliser shouldn't expect to have access to a whole lemma at once, and could potentially improve performance in some cases.

loiclec (changes requested):
Very nice refactor, it is a lot simpler :) I made three comments:
1. one about about potential performance issues
2. asking for clarification about the `char_map` and the  inputs that `normalize_str` expects
3. one about a potential bug in the default implementation of `normalize`

I also get a few warnings when compiling charabia now:
```
warning: unused import: `NormalizerOption`
 --> src/normalizer/chinese.rs:5:25
warning: unused import: `NormalizerOption`
 --> src/normalizer/control_char.rs:3:25
warning: unused import: `NormalizerOption`
 --> src/normalizer/latin.rs:5:25
warning: unused import: `NormalizerOption`
 --> src/normalizer/nonspacing_mark.rs:6:25
```

src/normalizer/mod.rs
```diff
@@ -78,17 +72,51 @@ pub trait Normalizer: Sync + Send {
...
+                    token.lemma = Cow::Owned(lemma);
+                    token.char_map = Some(char_map);
+                }
+                None => {
+                    let mut buffer = [0; 4];
+                    let mut char_map = Vec::new();
+                    let mut lemma = String::new();
+                    for c in token.lemma().chars() {
+                        let char_str = c.encode_utf8(&mut buffer);
+                        let normalized = self.normalize_str(char_str);
```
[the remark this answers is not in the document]
ManyTheFish: so far, performance is not my priority, I want to simplify the introduction of a new normalizer.
In the related issue, I was considering adding a `normalize_char` function in addition to `normalize_str`, then adding a default implementation of `normalize_str` that calls `normalize_char`.
I saw two issues with doing this:
1) We have more methods to implement because calling `normalize_char` in `normalize_str` is not always the best approach. For instance when your normalizer removes characters.
2) All the used libraries don't implement a function that takes a char in parameters

However, you're right we may lose performance due to allocations 🤔
I may find a way to do it by implementing several traits like `CharNormalizer` and `StrNormalizer` with a default implementation of the trait `Normalizer` for each trait, is it at least possible? 🤔

src/normalizer/lowercase.rs
```diff
@@ -1,33 +1,32 @@
...
+    fn normalize_str<'o>(&self, src: &'o str) -> Cow<'o, str> {
+        Cow::Owned(src.to_lowercase())
     }
 
-    fn should_normalize(&self, script: Script, _language: Option<Language>) -> bool {
+    fn should_normalize(&self, token: &Token) -> bool {
         // https://en.wikipedia.org/wiki/Letter_case#Capitalisation
-        matches!(script, Script::Latin | Script::Cyrillic | Script::Greek | Script::Georgian)
+        matches!(token.script, Script::Latin | Script::Cyrillic | Script::Greek | Script::Georgian)
+            && token.lemma.chars().any(char::is_uppercase)
```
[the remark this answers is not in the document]
ManyTheFish: Yes, It was done in the `normalize` method before, so we had different checks at different positions. It was a bit inconsistent so I chose to mutualize the starting condition.

src/normalizer/mod.rs
ManyTheFish: > I wonder, however, what a "character" from the char_map is. Is it always a single unicode code point (a char in Rust) or can it be a grapheme cluster?

It's always a char initially. Because in Meilisearch we work with normalized words but we want to highlight original words, the goal of the char map is to retrieve the original character's "position" in order to highlight the original word properly based on the number of characters that should be highlighted in the normalized version.

> In either case, I think there is a risk of misunderstanding about the kind of inputs that normalise_str accepts. I initially expected it to accept whole strings such as lion, but it seems like it is mostly given single characters unless options.create_char_map is false.

It's not always the case, the method normalize is overridden in the japanese and the lowercase normalizers skip the char map creation and call `normalize_str` over the whole string.

> If a character in the char_map is guaranteed to be a char, then maybe it would be worth having a normalize_char trait function? It could clarify that a normaliser shouldn't expect to have access to a whole lemma at once, and could potentially improve performance in some cases.

Yes but no, "yes" for the first normalizer that creates a char map, then "no" because the next normalizers may receive more than one character corresponding to the original character.
For instance, `deunicode` transforms 🦄 into "unicorn" and then the next normalizer will receive the whole "unicorn" string corresponding to the original character 🦄. It's not the best example I found sorry 😬

src/normalizer/mod.rs
```diff
@@ -78,17 +72,51 @@ pub trait Normalizer: Sync + Send {
...
+                    }
+
+                    token.lemma = Cow::Owned(lemma);
+                    token.char_map = Some(char_map);
+                }
+            }
+        } else if let Cow::Owned(lemma) = self.normalize_str(token.lemma()) {
+            token.lemma = Cow::Owned(lemma);
+        }
+
```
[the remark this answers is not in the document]
ManyTheFish: you're right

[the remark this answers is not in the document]
ManyTheFish: fixed

src/normalizer/mod.rs
```diff
@@ -78,17 +72,127 @@ pub trait Normalizer: Sync + Send {
...
-    fn normalize<'o>(
-        &self,
-        token: Token<'o>,
-        options: NormalizerOption,
-    ) -> Box<dyn Iterator<Item = Token<'o>> + 'o>;
+    fn normalize<'o>(&self, token: Token<'o>, options: NormalizerOption) -> Token<'o>;
+
+    /// Return true if the normalizer can process Token of a specific [`Script`] and [`Language`].
+    ///
+    /// Some normalizer are specialized for a `Script` or/and a `Language` and shouldn't be called on every `Token`s.
```
loiclec, line 79: How does one write a normalizer where `should_normalize` is not called on every token?

src/normalizer/mod.rs
```diff
@@ -78,17 +72,127 @@ pub trait Normalizer: Sync + Send {
...
+    match s {
+        Cow::Borrowed(s) => Cow::Borrowed(&s[..new_size]),
+        Cow::Owned(s) => Cow::Owned(s[..new_size].to_string()),
+    }
+}
+
+pub trait CharNormalizer: Sync + Send {
+    fn normalize_char(&self, c: char) -> Option<CharOrStr>;
+
+    fn normalize_cow_str<'o>(&self, s: Cow<'o, str>) -> Cow<'o, str> {
```
loiclec, line 93: Why does `normalize_cow_str` take a `Cow` as argument, as opposed to a `&str`? If I understand correctly, we don't make use of the `Owned` variant at all. But could someone write an alternative implementation that makes use of it?

loiclec (commented):
Thank you for the changes :) 
I left two small comments again, but we can merge it if you want.

src/normalizer/mod.rs
ManyTheFish: No every normalizer are not called for every token, but, should_normalize is always called to know if a normalizer should be called

src/normalizer/mod.rs
ManyTheFish: > Why does normalize_cow_str take a Cow as argument, as opposed to a &str?  If I understand correctly, we don't make use of the Owned variant at all

for a token, you can have several normalizers called in a row, so for the first normalizer call, you'll always have a borrowed value, but, this first normalizer could modify the borrowed value, for instance, a lowercase, and creates an Owned value. Then, the next normalizer will have an owned value as a parameter.
Each normalizer will replace the `lemma`, and it is important to note that the previous version will be dropped. So let's play some possible scenarios:
1) `normalizer(1)` returns a `Borrowed(a)` version of `text` then `normalizer(2)` returns a `Borrowed(b)` version of `Borrowed(a)` which is a borrow of `text`, `Borrowed(a)` is dropped, everything is ok.
2) `normalizer(1)` returns an `Owned(a)` version of `text` then `normalizer(2)` returns a `Borrowed(b)` version of `Owned(a)` which is a borrow on a modified clone of `text`, `Owned(a)` is dropped, `Borrowed(b)` can't be valid because of it has a reference on `Owned(a)`.

These cases, force us to know if the passed value is borrowed or not to know if we have to clone it.
However, it's possible that my implementation is not the best, I can just say that it solves this issue. 😬 

> But could someone write an alternative implementation that makes use of it?

Yes but I'll never suggest it

loiclec (approved):

PR #162, opened 2022-11-12
Normalize Chinese by Z, Simplified, Semantic, Old, and Wrong variants

### LINKED ISSUE #144: Enhance Chinese normalizer by unifying `Z` `Simplified` and `Semantic` variants
Following [the official discussion about Chinese support](https://github.com/meilisearch/product/discussions/503) in Meilisearch, it is relevant to normalize Chinese characters by unifying `Z` `Simplified` and `Semantic` variants before transliterating them into Pinyin.
> to know more about each variant, you can read the [dedicated report on unicode.org](http://www.unicode.org/reports/tr38/#N10211)

There are several dictionaries listing variations that we can use, I suggest using the [kvariants](https://github.com/hfhchan/irg/blob/master/kVariants.txt) dictionary made by hfhchan (see the [related documentation on the same repo](https://github.com/hfhchan/irg/blob/master/kVariants.md)).

### technical approach
Import and Rework the dictionary to be a key-value binding of each variant, then, in the Chinese normalizer, convert the provided character before [transliterating it into Pinyin](https://github.com/meilisearch/charabia/blob/1244b9d2ceef89c907c1b214c924ce7c81ee433b/src/normalizer/chinese.rs#L24). 

- [/src/normalizer/chinese.rs](https://github.com/meilisearch/charabia/blob/main/src/normalizer/chinese.rs)

### Misc
related to https://github.com/meilisearch/product/discussions/503

> Hey! 👋 

choznerol:
# Pull Request

## Related issue
Fixes #144

## What does this PR do?

As titled, use [`kVariants.txt`](https://github.com/hfhchan/irg/blob/master/kVariants.md) as a dictionary to enhance Chinese normalization.

### 1. Also normalizing `old` and `wrong!` variants

> .. , it is relevant to normalize Chinese characters by unifying `Z` `Simplified` and `Semantic` variants before transliterating them into Pinyin.

There are also `old` and `wrong!` variants in [`kVariants.txt`](https://github.com/hfhchan/irg/blob/master/kVariants.md#format). I didn't see a reason not also to handle them, so they are also convert.

### 2. Confirm direction of conversion

For `=` `old`, `sem`, `wrong!` variants, I think it's obvious we want to convert from Source Ideograph to Destination Ideograph. However, for `simp` I personally think the same but am not 100% sure if there would be other considerations. The reason I think traditional variants should be the normalized form includes:
1. Traditional variants seem to be the source of truth, just like Source Destination in `=` `old`, `sem` and `wrong!` all represent source of truth.
2. A log of simplified variants seems to be rendered unsuccessfully (the boxes of Unicode codepoint like 𧦛). I would worry if `ToPinyin` could be handled these simplified variants correctly if they are chosen as normalized form. <br/> image

If there is a preferred way to improve vendoring the dictionary (e.g. create a crate for this?), I'd love to look into it, but probably in a separate follow-up PR.

## PR checklist


src/normalizer/chinese.rs
```diff
@@ -4,6 +4,9 @@ use super::CharNormalizer;
 use crate::detection::{Language, Script};
 use crate::normalizer::CharOrStr;
 use crate::Token;
+use kvariants::KVARIANTS;
+
+mod kvariants;
 
 /// Normalize Chinese characters by converting them into Pinyin characters.
```
ManyTheFish, line 11: We should change this documentation

ManyTheFish (changes requested):

choznerol:
> I tried to convert every codepoint of the dictionary into Pinyin to see in which proportions we managed to have a conversion. Most of the time, the pinyin normalizer manages to convert the output variants, but, sometimes, the output has no conversion where the input has.
> If we want to maximize the Pinyin conversion, then I suggest changing the normalizer into:
> 
> However, if you think that maximizing the Pinyin conversion is a bad idea, let me know.

I also tried the `to_pinyin()` conversion in fae1c05 for both the source ideograph and destination ideograph. The result can be found on [demo_(simplified|all)_kvariant_to_pinyin_converiton](https://docs.google.com/spreadsheets/d/1OA9lIDn5STX4UOKDvYfhjcjs1vKXu1Ha84-ER5jGrek/edit?usp=sharing):

demo_simplified_kvariant_to_pinyin_converiton

Here is my interpretation of [demo_simplified_kvariant_to_pinyin_converiton](https://docs.google.com/spreadsheets/d/1OA9lIDn5STX4UOKDvYfhjcjs1vKXu1Ha84-ER5jGrek/edit#gid=693606809):
- 2541 `SAME_PINYIN`s: Most kvariant normalization actually has no effect on the final pinyin.
- 1053 `SOURCE_NO_PINYIN`s: These simplified variants can now benefit from the kvariant normalization. That's great.
- 211 `BOTH_NO_PINYIN`s: The kvariant normalization has no effect on the final result. There's nothing we can do here.
- 156 `DIFFERENT_PINYIN`s: By roughly browse through the cases I can recognize, I can't easily tell which one always results in a better pinyin. Sometimes it's just due to lack of context, for example, [辗/輾](https://dict.revised.moe.edu.tw/dictView.jsp?ID=3106&la=0&powerMode=0) can be either [zhǎn](https://dict.revised.moe.edu.tw/search.jsp?word=%E8%BC%BE%20%E3%84%93%E3%84%A2%CB%87) or [niǎn](https://dict.revised.moe.edu.tw/search.jsp?word=%E8%BC%BE%20%E3%84%8B%E3%84%A7%E3%84%A2%CB%87) depending on whether it means "toss and turn" or "rolling" in the current context.
- 8 `DESTINATION_NO_PINYIN`s: Your snippet from `2.` of https://github.com/meilisearch/charabia/pull/162#issuecomment-1313700254 can fix this!

This interpretation also applys to [demo_all_kvariant_to_pinyin_converiton]](https://docs.google.com/spreadsheets/d/1OA9lIDn5STX4UOKDvYfhjcjs1vKXu1Ha84-ER5jGrek/edit#gid=1341790256)

src/normalizer/chinese.rs
choznerol: Added b25c6d37d1e9b4dcf435c243dc1155cf28fcaf0a

choznerol:
> ### 3. Alternatives to copying and embedding the dictionary
> 
> > If there is a preferred way to improve vendoring the dictionary (e.g. create a crate for this?), I'd love to look into it, but probably in a separate follow-up PR.
> 
> Yes! I'd prefer to have a dictionary keeping only the useful data in order to reduce the crate size. So converting each line from `釠 (U+91E0) wrong! 亂 (U+4E82)` to something like `釠,wrong!,亂` or even a more compressed form would be better. Your suggestion of creating a separate crate is a good idea, therefore, we already have several dictionaries that have been manually converted and pushed in raw in Charabia like the [nonspacing-marks](https://github.com/meilisearch/charabia/blob/3dde30fb75f381ec3155d853699278eddd1934bc/src/normalizer/nonspacing_mark.rs#L10). If you are interested in creating it, I just request to do it in a separate PR! 😬

Ah, I see, that makes sense! I am interested in addressing this. Will do in another follow-up PR.

ManyTheFish (approved):
Well, what an excellent job!
Let's merge it. Then, if you want to dig deeper into creating an isolated crate for the dictionaries, I let you open a new PR for it, and I'll create a new crate name in consequence.

Thanks again for your work on this!

Bors merge

src/normalizer/chinese/kvariants.rs
```diff
@@ -0,0 +1,145 @@
...
+
+pub static KVARIANTS: Lazy<HashMap<char, KVariant>> = Lazy::new(|| {
+    // The tab separated format is like:
+    //
+    //   㨲 (U+3A32)	wrong!	㩍 (U+3A4D)
+    //   铿 (U+94FF)	simp	鏗 (U+93D7)
+    //   㓻 (U+34FB)	sem	    剛 (U+525B)
+    //   ...
+    //
+    let file = fs::File::open("dictionaries/txt/chinese/kVariants.tsv").unwrap();
```
choznerol, line 37: Hi @ManyTheFish I didn't realize until now that I probably should have use `include_bytes!` instead of `File::open` here. Not sure if this will break when packaged and released. I'm working on a follow-up PR to fix it.

ManyTheFish: Hello @choznerol, nice! Thank you!

PR #169, opened 2022-12-06
Classify a Token before normalizing it avoiding to have false positive in stop words

### LINKED ISSUE #150: Classify tokens after Segmentation instead of Normalization
Classifying tokens after Segmentation instead of Normalization in the tokenization pipeline would enhance the precision of the stop_words classification.
Today, stop words need to be normalized to be properly classified, however, the normalization is more or less lossy and can classy unexpected stop words.
For instance, in French `maïs` (`corn` in 🇬🇧) would be normalized as `mais` (`but` in 🇬🇧), and so, `maïs` will be classified as a stop word if the stop word list contains `mais`.
This would not be possible if the classifier is called before the normalizer.

### Technical approach

Invert the [normalization step and the classification step in the tokenization process](https://github.com/meilisearch/charabia/blob/6d0a01d371088283d4e9bbadba24c0a0d2344571/src/tokenizer.rs#L24-L120)

ManyTheFish:
This change needed a refactor in order to better control lifetimes in the Library.

Fixes #150

### DIFF
--- a/src/classifier.rs
+++ b/src/classifier.rs
@@ -1,16 +1,17 @@
 use deunicode::deunicode_char;
 use fst::Set;
 
+use crate::segmenter::SegmentedTokenIter;
 use crate::token::SeparatorKind;
 use crate::{Token, TokenKind};
 
 /// Iterator over classified [`Token`]s.
-pub struct ClassifiedTokenIter<'o, 'sw, A> {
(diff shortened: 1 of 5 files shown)


Kerollmops (commented):
I like what you've done 🎉 
However, have you tried to integrate this branch into milli? I would prefer you try to make it work before merging all those lifetime changes.

Kerollmops (approved):
bors merge

PR #170, opened 2022-12-13
Use cargo workspace

choznerol:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/product/discussions/582 (part 1)

## What does this PR do?

Convert the codebase structure to [cargo workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html) according to the conclusion in https://github.com/meilisearch/product/discussions/582#discussioncomment-4331356 .

I found the changing related to cargo workspace can accumulate merge conflict pretty easily, so I'm opening this as a standalone PR. The actual work to bring in https://github.com/choznerol/kvariants will be another follow-up PR #171 base on this branch.

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

Thank you so much for contributing to Meilisearch!

PR #171, opened 2022-12-13
Use compressed kvariants to reduce crate size

choznerol:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/product/discussions/582 (part 2)

## What does this PR do?
~Based on #170 (merged),~ brining in [choznerol/kvariants] according to the conclusion of https://github.com/meilisearch/product/discussions/582#discussioncomment-4331356 .

#### 1. Vendor the dictionary file instead of fetching on demand
In [choznerol/kvariants] I fetch the `kVariants.tsv` only before build ([`ci.yml`](https://github.com/choznerol/kvariants/blob/a0d0826dbbdad1365516f640189979a40e81574a/.github/workflows/ci.yml#L19-L20)). However, since the `charabia` already vendored some dictionaries, I believe vendoring is preferred compare to rely on a successful `curl ...` during CI.

#### 2. Some minor fix
Just in case you have reviewed [choznerol/kvariants] before, I copied over the code as is and do the minor fixes (e.g. clippy) here in this PR.

## PR checklist

[choznerol/kvariants]: https://github.com/choznerol/kvariants


kvariants/build.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(Deserialize)]
+pub struct TsvRow {
+    lhs: String,
+    relation: String,
+    rhs: String,
+}
+
+fn main() {
+    // Tell Cargo that if the given file changes, to rerun this build script.
+    println!("cargo:rerun-if-changed=dictionaries/source/kVariants.tsv");
```
ManyTheFish, line 13: It may be interesting to add these 2 additional checks.
```suggestion
    println!("cargo:rerun-if-changed=dictionaries/source/kVariants.tsv");
    println!("cargo:rerun-if-changed=build.rs");
    println!("cargo:rerun-if-changed=Cargo.toml");
```

ManyTheFish (changes requested):

kvariants/build.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    compress_kvariant_txt().unwrap();
+}
+
+fn compress_kvariant_txt() -> Result<(), Box<dyn Error>> {
+    let reader = csv::ReaderBuilder::new()
+        .delimiter(b'\t')
+        .has_headers(false)
+        .from_path("dictionaries/source/kVariants.tsv");
```
ManyTheFish, line 22: We should have an absolute path.
```suggestion
    let manifest_dir = env::var_os("CARGO_MANIFEST_DIR").unwrap();
    let src_path = Path::from(manifest_dir)
        .join("dictionaries")
        .join("source")
        .join("kVariants.tsv");
    let reader = csv::ReaderBuilder::new()
        .delimiter(b'\t')
        .has_headers(false)
        .from_path(src_path);
```

kvariants/build.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let mut writer = csv::Writer::from_path("dictionaries/compressed/kVariants.csv")?;
```
ManyTheFish, line 24: We may store the compressed dictionary in the target dir instead, which will avoid keeping an empty `compressed` directory in the repository.
```suggestion
    let out_dir = env::var_os("OUT_DIR").unwrap();
    let dst_path = Path::from(out_dir).join("kVariants.csv");
    let mut writer = csv::Writer::from_path(dst_path)?;
```

kvariants/src/lib.rs
```diff
@@ -34,15 +33,15 @@ pub static KVARIANTS: Lazy<HashMap<char, KVariant>> = Lazy::new(|| {
     //   㓻 (U+34FB)	sem	    剛 (U+525B)
     //   ...
     //
-    let tsv = include_str!("../../../../dictionaries/txt/chinese/kVariants.tsv");
+    let dictionary = include_str!("../dictionaries/compressed/kVariants.csv");
```
ManyTheFish, line 36: Then we retrieve the dictionary directly from the out dir.
```suggestion
    let dictionary = include_bytes!(concat!(env!("OUT_DIR"), "/kVariants.csv"));
```

kvariants/src/lib.rs
```diff
@@ -34,15 +33,15 @@ pub static KVARIANTS: Lazy<HashMap<char, KVariant>> = Lazy::new(|| {
...
     let mut reader =
-        csv::ReaderBuilder::new().delimiter(b'\t').has_headers(false).from_reader(tsv.as_bytes());
+        csv::ReaderBuilder::new().has_headers(false).from_reader(dictionary.as_bytes());
```
ManyTheFish, line 38: using `include_bytes!` instead of `include_str!` upper should allow us to remove this:
```suggestion
        csv::ReaderBuilder::new().has_headers(false).from_reader(dictionary);
```

kvariants/bin/clean
```diff
@@ -0,0 +1,7 @@
+set -ex
```
ManyTheFish, line 1: could you add a `.sh` extension to the file and a shebang on the top of it:
```suggestion
#!/bin/sh

set -ex
```

kvariants/bin/sync_dictionaries
```diff
@@ -0,0 +1,9 @@
+set -ex
```
ManyTheFish, line 1: could you add a `.sh` extension to the file and a shebang on the top of it:
```suggestion
#!/bin/sh

set -ex
```

ManyTheFish (changes requested):

kvariants/build.rs
choznerol: Fixed in cd6c975

kvariants/build.rs
choznerol: Thanks! I wasn't aware of `CARGO_MANIFEST_DIR`. Fixed in d5413b1.

kvariants/build.rs
choznerol: Thanks! I wasn't aware of `OUT_DIR`. Fixed in 2c326cd.

kvariants/src/lib.rs
choznerol: Also fixed in 2c326cd.

kvariants/src/lib.rs
choznerol: Fixed in b696feb. The `let dictionary: &[u8]` annotation was added otherwise the `Read` trait bound wouldn't work for `&[u8; 137909]` from `include_bytes!`.

kvariants/bin/sync_dictionaries
choznerol: Also fixed in 6b80181

choznerol:
The current CI failure would be addressed in #177 instead.

ManyTheFish (approved):
Looks good to me, I let you rebase after your previous PR, and then I'll merge this one.

Thanks

choznerol:
> Looks good to me, I let you rebase after your previous PR, and then I'll merge this one.
> 
> Thanks

Thanks for the review! Rebased without manually conflict resolution. Only d3dd22b is new for CI to pass.

ManyTheFish (approved):
Nice!
Thanks a lot for your investment!

bors merge

PR #173, opened 2022-12-15
Update Cargo ignored files

ManyTheFish:
Change cargo ignored files allowing the crate to be released on crates.io


curquiza (approved):
I understand the purpose, and I trust you for this

PR #180, opened 2023-02-01
Fix script lang serialization

ManyTheFish:
The script serialization was broken because we merged Mandarin, Hiragana, and Katakana scripts into one unique Cj script.


Kerollmops (approved):
Nice, thank you bra'!
bors merge

PR #181, opened 2023-02-01
Split camelCase in Latin segmenter

### LINKED ISSUE #129: Latin script: Segmenter should split camelCased words
Today, Meilisearch is splitting `snake_case`, `SCREAMING_CASE` and `kebab-case` properly but doesn't split `PascalCase` nor `camelCase`.

### drawback
 Meilisearch doesn't completely support code documentation.

### enhancement
Make Latin Segmenter split camelCased/PascalCase words:
- `"camelCase"` -> `["camel", "Case"]`
- `"PascalCase"` -> `["Pascal", "Case"]`
- `"IJsland"` -> `["IJsland"]` (Language trap)
- `"CASE"` -> `["CASE"]` (another trap)

- [/src/segmenter/latin.rs](https://github.com/meilisearch/charabia/blob/dac6b422e43b666cc232430533e61a50abdd3341/src/segmenter/latin.rs)

> Hey! 👋 

goodhoko:
## Related issue
#129

## What does this PR do?

To improve recall and to be consistent with snake_case and kebab-case splitting that's already in place, make the Latin Segmenter split words on camelCase boundaries.

Define camelCase boundary as a lowercase letter directly followed by an uppercase one. (The position between them, to be precise.) This treats most cases and avoids the common pitfalls like eg. ALL_CAPS. What is not handled though, are abbreviations within a longer word. Especially in code it's common to write eg. "meiliAPIClient". With this implementation it's split into ["meili", "APIClient"]. Let me know if that's a blocker.

Leverage the Unicode General Categories
https://en.wikipedia.org/wiki/Unicode_character_property#General_Category
and their support in the Regex crate for matching lowercase and uppercase letters.

Put the logic into a separate module and expose API similar to UnicodeSegmentation's to keep the call-site in latin.rs clean and concise.

Fixes #129.

## Performance
I benchmarked this locally on my Apple-silicon MB Air with `cargo bench` and saw some significant regressions. The  `Latin/Fra` and `Latin/Eng` (in both `132` and `363` variants) regressed some 40% compared to a3eab3067742ab5a0f5570b6fcc242808883a95c. (Here's the full [report](https://github.com/meilisearch/charabia/files/10559955/report.zip)) I'm not sure how big of a problem this is. There definitely are many optimization opportunities but I wanted to propose the simplest solution first and eventually iterate from there.

---

PS: This is my very first rust contribution. Sorry for missing any conventions or leaving in rough edges.

### DIFF
--- a/charabia/src/segmenter/camel_case.rs
+++ b/charabia/src/segmenter/camel_case.rs
@@ -24,7 +24,9 @@ impl CamelCaseSegmentation for str {
     }
 }
 
-static CAMEL_CASE_BOUNDARY_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"\p{Ll}\p{Lu}").unwrap());
+/// Matches a lowercase letter followed by an uppercase one with any number of optional non-spacing marks in between.
+static CAMEL_CASE_BOUNDARY_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"\p{Ll}\p{Mn}*?\p{Lu}").unwrap());
 
 impl<'t> Iterator for CamelCaseParts<'t> {
     type Item = &'t str;
@@ -33,7 +35,7 @@ impl<'t> Iterator for CamelCaseParts<'t> {
         match self.state {
             State::Exhausted => None,
             State::InProgress { remainder } => {
-                // CamelCase boundary consists of 2 code-points. Avoid expensive regex evaluation on shorter strings.
+                // CamelCase boundary consists of at least 2 code-points. Avoid expensive regex evaluation on shorter strings.
                 // Note that using `remainder.chars().count() == 1` may catch more cases (non-ASCII strings)
                 // but the main focus here is on " ", "-" and similar that are abundantly produced
                 // by `split_word_bounds()` in the Latin segmenter and mere `len()` performs better at that.
@@ -44,10 +46,10 @@ impl<'t> Iterator for CamelCaseParts<'t> {
 
                 match CAMEL_CASE_BOUNDARY_REGEX.find(remainder) {
                     Some(mat) => {
-                        // By the nature of the regex, the match must contain exactly two chars and this should never panic.
-                        let lowercase_letter_length =
-                            mat.as_str().chars().next().unwrap().len_utf8();
-                        let boundary = mat.start() + lowercase_letter_length;
+                        // By the nature of the regex, the match must contain at least two chars and this should never panic.
+                        let uppercase_letter_length =
+                            mat.as_str().chars().last().unwrap().len_utf8();
+                        let boundary = mat.end() - uppercase_letter_length;
 
                         self.state = State::InProgress { remainder: &remainder[boundary..] };
                         Some(&remainder[..boundary])
@@ -82,4 +84,9 @@ mod test {
     test_segmentation!("SCREAMING", ["SCREAMING"], all_caps_is_not_split);
     test_segmentation!("resuméWriter", ["resumé", "Writer"], non_ascii_boundary_on_left);
     test_segmentation!("KarelČapek", ["Karel", "Čapek"], non_ascii_boundary_on_right);
+    test_segmentation!(
+        "resume\u{0301}Writer",
+        ["resume\u{0301}", "Writer"],
+        non_spacing_marks_are_respected
+    );
 }
--- a/charabia/src/segmenter/camel_case.rs
+++ b/charabia/src/segmenter/camel_case.rs
@@ -36,6 +36,15 @@ impl<'t> Iterator for CamelCaseParts<'t> {
         match self.state {
             State::Exhausted => None,
             State::InProgress { remainder } => {
+                // CamelCase boundary consists of 2 code-points. Avoid expensive regex evaluation on shorter strings.
+                // Note that using `remainder.chars().count() == 1` may catch more cases (non-ASCII strings)
+                // but the main focus here is on " ", "-" and similar that are abundantly produced
+                // by `split_word_bounds()` in the Latin segmenter and mere `len()` performs better at that.
+                if remainder.len() == 1 {
+                    self.state = State::Exhausted;
+                    return Some(remainder);
+                }
+
                 match CAMEL_CASE_BOUNDARY_REGEX.captures(remainder) {
                     Some(captures) => {
                         // By the nature of the regex, this group is always present and this should never panic.
--- a/charabia/src/segmenter/camel_case.rs
+++ b/charabia/src/segmenter/camel_case.rs
@@ -63,3 +63,23 @@ impl<'t> Iterator for CamelCaseParts<'t> {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::CamelCaseSegmentation;
+
+    macro_rules! test_segmentation {
+        ($text:expr, $segmented:expr, $name:ident) => {
+            #[test]
+            fn $name() {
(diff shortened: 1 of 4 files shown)


ManyTheFish:
Hello @goodhoko,
Well, 40% regression is too high, moreover, the Latin segmenter is the most used one, so I can't accept this PR at this step. 😔

I'll encourage you into suggesting a more performant implementation.

In any case, thank you for your investigation of this issue! ☺️

goodhoko:
@ManyTheFish Understood.

Before I try to optimize this, it would be good to know if splitting the abbreviations correctly (see the PR description) is a hard requirement.

Also, I'm afraid we can't really implement this without noticeable regression since we can't avoid scanning every word for camelCase boundaries. I totally agree that 40% is way too much. Let's see if I can squash that lower. I have a few ideas. .)

ManyTheFish:
Yes, sorry, about camelCase splitting abbreviations, I think it's ok like this, moreover, depending on the developer, abbreviations are considered as words, and so, `meiliAPIClient` could be spelled `meiliApiClient`.
Therefore, let's wait for an issue asking to change it. ☺️

goodhoko:
@ManyTheFish I just pushed two additional commits that combined bring the regression down to about 5 % (as measured on my laptop).

I also realized one more limitation of this implementation: It is not grapheme-cluster aware. For instance, it doesn't split `resuméWritter` where `é` is encoded as `['e', '\u{0301}']`. Though, it splits `resuméWriter` where `é` is just `'é'`.

ManyTheFish (approved):
> I also realized one more limitation of this implementation: It is not grapheme-cluster aware. For instance, it doesn't split resuméWritter where é is encoded as ['e', '\u{0301}']. Though, it splits resuméWriter where é is just 'é'.

So it would be the case for every [non-spacing marks](https://www.compart.com/en/unicode/category/Mn). 🤔
I'm wondering if it's really useful to handle this, I may be wrong, but I feel that camelCase is 100% used in English where there are no non-spacing marks.
What do you think?

goodhoko:
@ManyTheFish The failed formatting check reports errors in files this PR didn't change. I rebased (and force pushed) hoping it will resolve itself.

ManyTheFish:
Hello @goodhoko,

your PR looks good to me! A last thing, could you add a feature flag "latin-camelcase" adding it as a default feature?

Thanks a lot!

goodhoko:
@ManyTheFish Thanks!

Sorry to force-push but I was hitting some unstable features problems with the lindera dependency and I hoped a rebase would help. (updating my toolchain helped after all)

Anyway, I added just one more [commit](https://github.com/meilisearch/charabia/pull/181/commits/e25fc5d93a10af46d0c97d2478c84b95cc9148fb) that adds the feature. See the commit message for details.

ManyTheFish:
Hey @goodhoko, thank you for your work!

bors merge

ManyTheFish (approved):
bors merge

charabia/src/segmenter/camel_case.rs
```diff
@@ -0,0 +1,92 @@
...
+
+enum State<'t> {
+    InProgress { remainder: &'t str },
+    Exhausted,
+}
+
+impl CamelCaseSegmentation for str {
+    fn split_camel_case_bounds(&self) -> CamelCaseParts {
+        CamelCaseParts { state: State::InProgress { remainder: self } }
+    }
```
strohel, line 24: An alternative to implementing the iterator manually would be to use some of the existing iterator utilities, perhaps [`iter::from_fn()`](https://doc.rust-lang.org/std/iter/fn.from_fn.html):
```rust
fn split_camel_case_bounds(&self) -> impl Iterator<Item = &str> + '_ {
    let mut remainder = self;

    iter::from_fn(move || {
        if reminder.is_empty() {
            return None;
        }
        
        match reminder.find(...)
            ...
            return Some(remaidner[..boundary]);
    })
}
```

I have not tried, but this simplified example compiles: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c19ba3e384445a3c2cf054625134949f

**Caveat**: Rust currently doesn't allow "[Return position impl Trait in traits](https://rust-lang.github.io/impl-trait-initiative/RFCs/rpit-in-traits.html#return-position-impl-trait-in-traits-1)" (this is just a very draft RFC), so one cannot have the `-> impl Iterator<Item = &str> + '_` signature in `CamelCaseSegmentation`.

And I fear one cannot even use the specific [`FromFn`](https://doc.rust-lang.org/std/iter/struct.FromFn.html) struct as its generic closure argument `F` cannot be named.. :thinking:

strohel (commented):
Adding a post-mortem curiosity review, just for the exercise. :slightly_smiling_face:

charabia/src/segmenter/camel_case.rs
```diff
@@ -0,0 +1,92 @@
...
+                // CamelCase boundary consists of at least 2 code-points. Avoid expensive regex evaluation on shorter strings.
+                // Note that using `remainder.chars().count() == 1` may catch more cases (non-ASCII strings)
+                // but the main focus here is on " ", "-" and similar that are abundantly produced
+                // by `split_word_bounds()` in the Latin segmenter and mere `len()` performs better at that.
+                if remainder.len() == 1 {
+                    self.state = State::Exhausted;
+                    return Some(remainder);
+                }
+
+                match CAMEL_CASE_BOUNDARY_REGEX.find(remainder) {
```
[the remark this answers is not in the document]
goodhoko: I was expecting a suggestion to avoid Regex but not a complete working example! 🖤

Turns out, this yields a performance boost of about 5%. I submitted a [PR](https://github.com/meilisearch/charabia/pull/196) with this change. Hopefully the maintainers won't be mad that we nerd out in their repo. 😅

One question: What's the reason for explicit `move` on the closure? The compiler doesn't complain about its absence.

charabia/src/segmenter/camel_case.rs
goodhoko: I did try this over [here](https://github.com/goodhoko/charabia/commit/8f2173eaff6582aed5c745e58ed07bae3ed01013) and, to be honest, I think it doesn't really make the code any more simple or readable.

BTW: I know it's not the point of this suggestion, but using the bare remainder to track the state of the iterator is not merely enough. When the remainder is empty you may either want to yield and empty string or `None` depending on whether you were given an empty string to begin with or not. That's because splitting `""` should result in `[""]` and not `[]`.

charabia/src/segmenter/camel_case.rs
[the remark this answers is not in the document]
strohel: > One question: What's the reason for explicit `move` on the closure? The compiler doesn't complain about its absence.

I think I just copy-pasted it from a similar example, no reason to have it if compiler doesn't complain.

Without the `move`, the "desugared closure struct" (closures can be imagined as auto-generated structs with a call method) should look like
```rust
struct Closure {
    last_char_was_lower_case: bool  # mut is inherited, see https://stackoverflow.com/a/27023282/4345715
}
```
Without the `move` it would look like
```rust
struct Closure<'a> {
    last_char_was_lower_case: &'a mut bool
}
```

I.e. with the `move`, this closure becomes fully owned (and can be i.e. returned from function).

Without the `move`, the lifetime of the closure is limited by the `last_char_was_lower_case`'s lifetime. But that's apparently sufficient here.

charabia/src/segmenter/camel_case.rs
strohel: Right, I see. I didn't realize the complication with an empty string.

PR #182, opened 2023-02-06
feat: enable nonspacing-marks normalizer for Greek scripts and introduce Greek normalizer which unifies sigma character 

### LINKED ISSUE #178: Greek Script: Normalize accents
# Remove Greek nonspacing-marks

In Meilisearch v1.0.0, we already decompose characters from their nonspacing-marks, for example, `ώ (U+03CE)` is decomposed in `ω (U+03C9) + ◌́ (U+0301)`, but we don't remove the nonspacing-marks.

A nonspacing-marks remover exists in Meilisearch but isn't activated in this version.

ο = ό
ι = ί
η = ή
ε = έ
ω = ώ
υ = ύ

## Possible enhancement

Make a PR activating nonspacing-marks normalizer for Greek script:
https://github.com/meilisearch/charabia/blob/38ba8570c3b7531fd44bb8773328681fcd21a7f9/charabia/src/normalizer/nonspacing_mark.rs#L28-L31

## Related
- https://github.com/meilisearch/product/discussions/596

### LINKED ISSUE #179: Greek script: Normalize Sigma
# Normalize Sigma
Meilisearch v1.0.0 doesn't normalize final-sigma, and so, `ς` is considered different than `σ`.

> ς = σ

## Possible enhancement

Create a new normalizer converting `ς` into `σ`.

See the [Normalizer template](https://github.com/meilisearch/charabia/blob/main/charabia/src/normalizer/dummy_example.rs) to fix this issue.

## Related

- https://github.com/meilisearch/product/discussions/596

--------------------------------

> Hey! 👋 

cymruu:
# Pull Request
1. enable nonspacing-marks normalizer for Greek scripts
2. introduce Greek normalizer which unifies sigma character
## Related issue
Fixes #178 & Fixes #179

## PR checklist


ManyTheFish:
Hello,
Nice contribution!

Just a small request before merging your PR,
Could you add a dedicated greek benchmark to the list:
https://github.com/meilisearch/charabia/blob/main/charabia/benches/bench.rs#L7-L24

below is the Wikipedia article on Greece if you want to take a sentence from it:
https://el.wikipedia.org/wiki/%CE%95%CE%BB%CE%BB%CE%AC%CE%B4%CE%B1

Thank you for your contribution!

cymruu:
@ManyTheFish thank you for reviewing 😄 I've added the Greek normalizer benchmarks. 
Let me know if you have any more suggestions

ManyTheFish (dismissed):
Hello @cymruu, 
your PR can be merged at this step, nice work.

That's been said, if you want to go further in the implementation, we can think about some optimizations. 🤔
You've used the `CharNormalizer` which is a simple interface to create a normalizer, however, there are several drawbacks:
1) It iterates over all the characters of each greek word whereas the 'ς' can only be the last one
2) It creates a char_map for each greek word whereas it's never useful because the only character that changes is 'ς' into 'σ' which are both 2 bytes long. That always gives an identity char_map which is the same as no char map.

An enhancement could be, as the [Japanese normalizer](https://github.com/meilisearch/charabia/blob/main/charabia/src/normalizer/japanese.rs#L23-L45), to derive the `GreekNormalizer` directly from the trait `Normalizer` instead of `CharNormalizer`.
This will allow you to only check the last character using [`ends_with`](https://doc.rust-lang.org/std/string/struct.String.html#method.ends_with), moreover, you can skip easily the char map creation by only avoiding modifying it in the token.

Thanks!

cymruu:
@ManyTheFish Thanks for such an instructive comment. I really appreciate it.
I pushed a commit that improves the code.

charabia/src/normalizer/greek.rs
```diff
@@ -0,0 +1,70 @@
...
+///
+pub struct GreekNormalizer;
+
+impl Normalizer for GreekNormalizer {
+    // converting  "ς" to "σ" doesn't change the characters length,
+    // so the `normalize` method is overloaded to skip the useless char_map computing.
+    fn normalize<'o>(&self, mut token: Token<'o>, _options: NormalizerOption) -> Token<'o> {
+        if token.lemma.ends_with("ς") {
+            token.lemma = Cow::Owned(token.lemma.replace("ς", "σ"))
+        }
```
ManyTheFish, line 17: Sorry to suggest something else than I suggested before, but I think we can target and replace only the last character:
```suggestion
        if let Some(prefix) = token.lemma.strip_suffix('ς') {
            token.lemma = Cow::Owned([prefix, "σ"].concat())
        }
```

however, this implementation loses clarity, so don't hesitate to run `cargo bench` to compare the Greek tokenization, before committing the changes. 🤔

ManyTheFish (changes requested):
One last test, thank you for your work!

charabia/src/normalizer/greek.rs
cymruu: I've ran the benchmarks. Here are the results:

initial approach f5adfbd
<details>

image

</details>

1st optimalization d0701c3 - **performance has improved**
<details>

image

</details>

`.strip_prefix` approach - **segmentation performance has regressed** - does it make sense? Can a change in normalizer affect the process of segmentation? 
**tokenize/130/Greek/Ell benchmark** -2.0098% 
other tokenize benchmarks didn't report any change  
 
<details>

image

</details>

ManyTheFish: Well, you only worked on the Normalization which is not called during the segmentation, so let's ignore it.
The first optimization is good, skipping the char_map creation and checking only the last character works well.
About the second one, only the first benchmark shows differences, but we can easily find the reason by counting the number of words containing a ς, the first one has 4 occurrences for 130bytes whereas the second one has only 2 occurrences for 364bytes.
Let's keep the `strip_prefix` implementation, it seems to have an impact depending on the frequency of word ending by ς. 🤔

cymruu: >About the second one, only the first benchmark shows differences, but we can easily find the reason by counting the number of words containing a ς, the first one has 4 occurrences for 130bytes whereas the second one has only 2 occurrences for 364bytes.
Oh yes, that makes sense.
Maybe one of benchmarks strings should be changed, so the second benchmark case is not simpler to handle for normalizer? Then it might be easier to actually compare benchmark results. What do you think?

ManyTheFish (approved):
> Oh yes, that makes sense.
Maybe one of benchmarks strings should be changed, so the second benchmark case is not simpler to handle for normalizer? Then it might be easier to actually compare benchmark results. What do you think?

Well, I think that we shouldn't change the benchmarks to our advantage unless we can prove that the text is a representative greek sentence.

Let's keep it like this! Thank you so much for your commitment! 

Bors merge

Issue #151, opened 2022-10-10
Upgrade dependencies


curquiza:
Checked with @ManyTheFish 

We need to upgrade the dependencies of Charabia for v0.30.0

ManyTheFish:
@curquiza let's delay this task for the next release of Meilisearch, I don't plan to do it for v0.30.0

curquiza:
Hello @ManyTheFish! I suppose this is due to a lack of time during this sprint? Can you confirm since we want to be transparent? 😊

ManyTheFish:
Indeed, I can't be 100% sure about having the time to do it before the pre-release of v0.30.0, moreover, the Hacktoberfest is making a lot of changes in the code and in the dependencies, forcing me to do a final checkout before doing a potential release of Charabia.
I'd prefer to update the dependencies after this checkout.

curquiza:
Got it!
Thanks for letting us know!

curquiza:
@ManyTheFish could we do this for v1.1.0? Related to https://github.com/meilisearch/meilisearch/issues/3285 which is an issue for v1

ManyTheFish:
Yes, I'm doing it this week before releasing a new version of Charabia.

ManyTheFish:
Fixed by #209

PR #183, opened 2023-02-06
Upgrade Lindera and use compressed dicts

ManyTheFish:
Upgrade lindera to the last version and use the compressed dictionary reducing the size of a dummy binary from 209MB
to 44MB.

related to #151

### DIFF
--- a/charabia/src/segmenter/korean.rs
+++ b/charabia/src/segmenter/korean.rs
@@ -1,5 +1,6 @@
 use lindera::mode::{Mode, Penalty};
-use lindera::tokenizer::{DictionaryConfig, DictionaryKind, Tokenizer, TokenizerConfig};
+use lindera::tokenizer::{DictionaryConfig, Tokenizer, TokenizerConfig};
+use lindera::DictionaryKind;
 use once_cell::sync::Lazy;
 
 use crate::segmenter::Segmenter;
@@ -11,17 +12,19 @@ pub struct KoreanSegmenter;
 
 static LINDERA: Lazy<Tokenizer> = Lazy::new(|| {
     let config = TokenizerConfig {
-        dictionary: DictionaryConfig { kind: DictionaryKind::KoDic, path: None },
+        dictionary: DictionaryConfig { kind: Some(DictionaryKind::KoDic), path: None },
         mode: Mode::Decompose(Penalty::default()),
         ..TokenizerConfig::default()
     };
-    Tokenizer::with_config(config).unwrap()
+    Tokenizer::from_config(config).unwrap()
 });
 
 impl Segmenter for KoreanSegmenter {
     fn segment_str<'o>(&self, to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
         let segment_iterator = LINDERA.tokenize(to_segment).unwrap();
-        Box::new(segment_iterator.into_iter().map(|token| token.text))
+        Box::new(
+            segment_iterator.into_iter().map(|token| &to_segment[token.byte_start..token.byte_end]),
+        )
     }
 }
 
--- a/charabia/Cargo.toml
+++ b/charabia/Cargo.toml
@@ -22,7 +22,9 @@ serde = "1.0"
 slice-group-by = "0.3.0"
 unicode-segmentation = "1.6.0"
 whatlang = "0.16.1"
-lindera = { version = "=0.17.0", default-features = false, optional = true }
+lindera = { version = "=0.21.0", default-features = false, optional = true }
+lindera-ko-dic = { version = "0.21.0", default-features = false, optional = true }
+lindera-ipadic = { version = "0.21.0", default-features = false, optional = true }
 pinyin = { version = "0.9", default-features = false, features = [
   "with_tone",
 ], optional = true }
@@ -40,11 +42,11 @@ chinese = ["dep:pinyin", "dep:jieba-rs"]
 hebrew = []
 
 # allow japanese specialized tokenization
-japanese = ["lindera/ipadic"]
+japanese = ["lindera/ipadic", "lindera-ipadic/compress"]
 japanese-transliteration = ["dep:wana_kana"]
 
 # allow korean specialized tokenization
-korean = ["lindera/ko-dic"]
+korean = ["lindera/ko-dic", "lindera-ko-dic/compress"]
 
 # allow thai specialized tokenization
 thai = []
--- a/charabia/src/segmenter/japanese.rs
+++ b/charabia/src/segmenter/japanese.rs
@@ -1,5 +1,6 @@
 use lindera::mode::{Mode, Penalty};
-use lindera::tokenizer::{Tokenizer, TokenizerConfig};
+use lindera::tokenizer::{DictionaryConfig, Tokenizer, TokenizerConfig};
+use lindera::DictionaryKind;
 use once_cell::sync::Lazy;
 
 use crate::segmenter::Segmenter;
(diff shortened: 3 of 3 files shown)


charabia/src/segmenter/korean.rs
```diff
@@ -11,17 +12,19 @@ pub struct KoreanSegmenter;
...
-    Tokenizer::with_config(config).unwrap()
+    Tokenizer::from_config(config).unwrap()
 });
 
 impl Segmenter for KoreanSegmenter {
     fn segment_str<'o>(&self, to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
         let segment_iterator = LINDERA.tokenize(to_segment).unwrap();
-        Box::new(segment_iterator.into_iter().map(|token| token.text))
+        Box::new(
+            segment_iterator.into_iter().map(|token| &to_segment[token.byte_start..token.byte_end]),
```
ManyTheFish, line 26: @mosuka, I didn't manage to use the method [`get_text`](https://github.com/lindera-morphology/lindera/blob/51c7e2feb8b1282f24fbd5c018f360e622116965/lindera-core/src/token.rs#L65-L67) of Lindera, I put below the error:
```
  --> charabia/src/segmenter/japanese.rs:25:59
   |
25 |         Box::new(segment_iterator.into_iter().map(|token| token.get_text()))
   |                                                           ^^^^^^^^^^^^^^^^ returns a reference to data owned by the current function
   ```
   
It's probably due to a missing lifetime in the function implementation that explicitly makes the link between the original string and the token text. But I'm not sure. 🤔 

> note: This was working in a previous version (0.17) using directly `token.text` 

Anyway, I found a workaround, but I wanted to notify you about the limit of the function `get_text()`. ☺️

mosuka: @ManyTheFish 
Thank you for letting me know!

irevoire (approved):
:mouse:

ManyTheFish:
bors merge

charabia/src/segmenter/korean.rs
dureuill: :thinking: looking a bit into this, this is due to the `Token` type now containing a `text: Cow<'dict, str>` rather than a `text: &'original_data str` member.

Previously, the code could get away by copying the shared reference to the original data (`token.text` would create a new shared reference to the same data) because `token` had no ownership of this data, it was just borrowing it.

Now, the token may actually own the string inside of the `Cow` (when it is the `Owned` variant), and so `token.get_text()` cannot possibly outlive the token, because that could cause a Use-After-Free if the token was dropped before the `token.get_text()` reference, and so this is forbidden by safe Rust.

I lack context to evaluate its correctness, but I'm a bit concerned by the workaround: I assume that a `Cow` was used due to a need of sometimes modifying the originally borrowed data (I believe this is done in the `Token::set_text` function). Code that relies on indexing the original will miss the modifications performed on the token. No idea if this will cause an issue in practice, but from a type standpoint this doesn't look right.

Now, as to what we could do to fix the issue, I see multiple ways depending on the reason the `text` was wrapped in a `Cow`. 

1. Make it your caller's problem. Add a `Token::into_components(self) -> (Cow<'a, str>, ...) method that consumes the token and returns its fields. Thus the caller gets control over the possible allocation that was made in the `Cow::Owned` case and can act upon it. Of course, this means changing more code in the caller here as it is currently returning an iterator of borrowed strings, and will now need to account for the fact that it may own some of them, so return an iterator of Cows.
2. Make it your caller's problem, take 2. Use `Rc<RefCell<String>>` (or a more specialized shared string variant with the refcount stored inline), and have `get_text` return a clone of the `Rc`. Again, the string will be owned and no longer referencing any data elsewhere, so the caller's signature here will need to be changed to an iterator of these refcounted strings.
3. Revert the token type to a direct reference.

Note that adding the `'a` lifetime to self in `get_text` will express the constraint that the lifetime of the Token should be the same as the lifetime of the dictionary. Unfortunately, this is not going to help us here because this lifetime is still going to be limited by the scope of the lambda in `map`, and thus we will encounter the same error.

See playground for [some examples](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=edea5ce468b90ccbe81135b686aa2214).

Note that, except if we're reverting the `Token` to a direct reference, there is no way to fix the lifetime issue without changing the caller's code to take into account the change in ownership. As far as I know it is very difficult to abstract on ownership in Rust.

Hope this helps :-)

ManyTheFish: Hello @dureuill, nice to see you passing by! ☺️

About `get_text`, I totally agree with you, it's no more possible to have a reference over the original text. And so we don't want to rely on it.

> I lack context to evaluate its correctness, but I'm a bit concerned by the workaround of this PR: I assume that a Cow was used due to a need of sometimes modifying the originally borrowed data (I believe this is done in the Token::set_text function). Code that relies on indexing the original will miss the modifications performed on the token. No idea if this will cause an issue in practice, but from a type standpoint this doesn't look right.

I understand your concern, so let's give a bit of context. (@mosuka don't estate to correct me if I'm wrong)
Behind Lindera it's a Viterbi algorithm that runs, basically, relying on a weighted dictionary, we index every possible word that matches the text in "edges", then, based on the weights of each edge, we keep the edges chain that optimizes the most the weight cost.
That's been said, the creation of the edge is done in [Lattice::set_text](https://github.com/lindera-morphology/lindera/blob/51c7e2feb8b1282f24fbd5c018f360e622116965/lindera-core/src/viterbi.rs#L167-L255), in this function:
1) [we iterate over the text byte by byte and we create a sub-str named "suffix" containing the remaining part of the text](https://github.com/lindera-morphology/lindera/blob/51c7e2feb8b1282f24fbd5c018f360e622116965/lindera-core/src/viterbi.rs#L189-L196)
2) [we create an edge for each dictionary word that is a prefix of the suffix](https://github.com/lindera-morphology/lindera/blob/51c7e2feb8b1282f24fbd5c018f360e622116965/lindera-core/src/viterbi.rs#L225-L226)
  - note that the start_index and the stop_index refer to the string passed in parameters
3) [later the function calculate_path_cost will compute the cost of every path without changing start_index and stop_index](https://github.com/lindera-morphology/lindera/blob/51c7e2feb8b1282f24fbd5c018f360e622116965/lindera-core/src/viterbi.rs#L327-L353), honestly it's one of the more obscure parts for me but I didn't took the time to dig in it so far.
4) [at last we get the best path, keeping the start_index related to the original text](https://github.com/lindera-morphology/lindera/blob/51c7e2feb8b1282f24fbd5c018f360e622116965/lindera-core/src/viterbi.rs#L355-L371) which will be used to compute the boundaries used in Charabia

---------------------------------------

Now, why use a CoW in the last versions of Lindera? It seems that Lindera is no more only a Segmenter and provides a way to [Normalize and Filter tokens](https://github.com/lindera-morphology/lindera/search?q=set_text), for now, we are not using these processes in Charabia moreover, these processes don't seem to modify the `token.byte_start` nor `token.byte_end`. 🤔

dureuill: Wow, thank you that's a detailed answer with quite a bit of context indeed ❤️ 

> Now, why use a CoW in the last versions of Lindera? It seems that Lindera is no more only a Segmenter and provides a way to [Normalize and Filter tokens](https://github.com/lindera-morphology/lindera/search?q=set_text).
> For now, we are not using these processes in Charabia

This makes me confident that indexing from the original text like we do in this PR is the correct thing to do ☺️ 

thanks again

ManyTheFish: @mosuka,
if you want some help with the redesign, I can allocate a bit of time during the next Meilisearch prerelease in March.
We can plan a call or create an asynchronous discussion on GitHub to exchange about your needs and ours.
That could lead to an implementation that is profitable for both.

What do you think? 🙂

mosuka: @ManyTheFish 
Thank you, that’s reassuring. Don't mind my other project. Lindera's filter mechanism is an experimental feature.
If there is something in Lindera's code that needs to be fixed, please send us a pull request.
Also, I've released Lindera 0.22.0, and changed the Token text back to &str, so please test it if you have time.
Thanks. :smiley:

mosuka: @ManyTheFish 
I sent a PR for resolving this problem. Please check it ou.
https://github.com/meilisearch/charabia/pull/195

PR #189, opened 2023-02-14
fix cargo fmt

ManyTheFish:


irevoire (approved):
noice :cactus: 
bors merge

PR #192, opened 2023-02-15
Regularly check for kVariant updates

### LINKED ISSUE #185: Create a CI keeping kvariants up-to-date
related to #184

goodhoko:
Related issue: #185

Add a workflow that checks if there are any updates to the kVariants dictionary mirrored from https://github.com/hfhchan/irg and in such case submit a PR.

Schedule it to run at 10:00 every Monday but also allow it to be triggered manually.

An example of a run can be seen [here](https://github.com/goodhoko/charabia/actions/runs/4188213705/jobs/7259060086) and the resulting (testing) PR over [here](https://github.com/goodhoko/charabia/pull/1) (note: the diff set is empty because the artificial change I introduced for testing has since been reverted).

Closes #185.

### DIFF
--- a/.github/workflows/update-kvariants.yml
+++ b/.github/workflows/update-kvariants.yml
@@ -1,0 +1,29 @@
+name: Check for kVariants updates
+
+on:
+  # Run this at 10:00 every monday.
+  schedule:
+    - cron:  '0 10 * * 1'
+  # Also allow triggering manually.
+  workflow_dispatch:
+
+env:
+  # Use a token that has write access to repo.
+  GH_TOKEN: ${{ secrets.MEILI_BOT_GH_PAT }}
+
+jobs:
+  check-for-kvariants-updates:
+    name: Check for kVariants updates
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Fetch latest version
+        run: ./kvariants/bin/sync_dictionaries.sh
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: Update kVariants dictionary
+          branch: update-kvariants
+          title: Update kVariants dictionary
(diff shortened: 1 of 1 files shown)


ManyTheFish:
Hello @goodhoko, so we have to add a CI env var named MEILI_BOT_GH_PAT,  that has the writing access on this repository, am I right?

goodhoko:
@ManyTheFish Sorry I should've mention this in the PR description. I took the variable from the [other workflow](https://github.com/meilisearch/charabia/blob/f653e77065d8f6a853a5be807c2786faef5951b9/.github/workflows/update-cargo-toml-version.yml#L13) so I assumed it could (and should) be used in this workflow as well.

Though, it's not strictly necessary to use a custom access token. If you enable the "Allow GitHub Actions to create and approve pull requests" option at the very bottom of the [Actions settings](https://github.com/meilisearch/charabia/settings/actions) the `GH_TOKEN` can be just left set to the default value.

ManyTheFish (approved):
I Approve this PR,
I request @curquiza to check the CI.

.github/workflows/update-kvariants.yml
```diff
@@ -0,0 +1,29 @@
+name: Check for kVariants updates
```
curquiza, line 1: It does not look like a CI check to me, but more a CI that updates the dictionaries.
I would also add a comment to describe the CI, like "Create a PR update the dictionaries if a changes has been found" (for example)

goodhoko: @curquiza Just pushed a [fixup commit](https://github.com/meilisearch/charabia/pull/192/commits/67a5e41af004dea31f1d2cb8315ad4b6ffd3ea49) that should address both of your remarks. 🙌

ManyTheFish (approved):
LGTM, thank you @goodhoko for your work, and thank you @curquiza for the double check!

bors merge

PR #194, opened 2023-02-16
move dictionnaries in charabia space

ManyTheFish:
To publish Charabia on crates.io, we need to move dictionaries in Charabia space


irevoire (approved):
Nice 🍥 

bors merge

PR #146, opened 2022-10-06
Add NonspacingMark normalizer

crudiedo:
# Pull Request

## Related issue
Fixes #137

## What does this PR do?
This PR adds new `NonspacingMarkNormalizer` that uses built-in `HashSet` data structure with the predefined set of marks (in a binary form) to remove nonspacing marks.  

It also removes the Hebrew normalizer (since the new one does the same job) and updates the Thai tests (since they're getting affected). 

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

Issue #137, opened 2022-09-20
Implement an efficient `Nonspacing Mark` Normalizer


ManyTheFish:
In the Information Retrieval (IR) context, removing Nonspacing Marks like diacritics is a good way to increase recall without losing much precision, like in Latin, Arabic, or Hebrew. 

## Technical Approach
Implement a  new Normalizer, named `NonspacingMarkNormalizer`, that removes [the nonspacing marks](https://www.compart.com/en/unicode/category/Mn) from a provided token (find a naive implementation with the exhaustive list in the `Misc` section).
Because there are a lot of sparse character ranges to match, it would be inefficient to create a big if-forest to know if a character is a nonspacing mark.
This way, I suggest trying several implementations of the naive implementation below in a small local project.

### Interesting Rust Crates
- [hyperfine](https://crates.io/crates/hyperfine): a small command-line tool to benchmark several binaries
- [roaring-rs](https://crates.io/crates/roaring): a bitmap data structure that has an efficient `contains` method
- [once_cell](https://crates.io/crates/once_cell): a good Library to create lazy statics already used in the repository

## Misc
- [naive implementation of `is_nonspacing_mark`](https://gist.github.com/ManyTheFish/2e05698d1c76c0504c3f7738de40c95d)
- related discussion about [the Arabic Language](https://github.com/meilisearch/product/discussions/139)

> Hey! 👋 
Before starting any implementation, make sure that you read the [CONTRIBUTING.md](https://github.com/meilisearch/charabia/blob/main/CONTRIBUTING.md#contributing) file.
In addition to the recurrent rules, you can find some guides to easily implement a `Segmenter` or a `Normalizer`.
Thanks a lot for your Contribution! 🤝

PR #196, opened 2023-02-27
Replace Regex with str::find

goodhoko:
@ManyTheFish Sorry to bring this up again but @strohel[1] had an excellent [suggestion](https://github.com/meilisearch/charabia/pull/181#discussion_r1117990017) to improve my previous PR which not only leaves the code subjectively nicer but improves performance by about 5%. I also added one more tiny simplification in the second commit.

See the commit messages for details.

---

[1] A friend of mine just so kind to give a throughout code review when I mention my first rust contribution to him. 🖤

### DIFF
--- a/charabia/src/segmenter/camel_case.rs
+++ b/charabia/src/segmenter/camel_case.rs
@@ -1,5 +1,4 @@
-use once_cell::sync::Lazy;
-use regex::Regex;
+use finl_unicode::categories::CharacterCategories;
 
 pub(crate) trait CamelCaseSegmentation {
     /// Returns an iterator over substrings of `self` separated on camelCase boundaries.
@@ -24,10 +23,6 @@ impl CamelCaseSegmentation for str {
     }
 }
 
-/// Matches a lowercase letter followed by an uppercase one with any number of optional non-spacing marks in between.
-static CAMEL_CASE_BOUNDARY_REGEX: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r"\p{Ll}\p{Mn}*?\p{Lu}").unwrap());
-
 impl<'t> Iterator for CamelCaseParts<'t> {
     type Item = &'t str;
 
@@ -35,22 +30,8 @@ impl<'t> Iterator for CamelCaseParts<'t> {
         match self.state {
             State::Exhausted => None,
             State::InProgress { remainder } => {
-                // CamelCase boundary consists of at least 2 code-points. Avoid expensive regex evaluation on shorter strings.
-                // Note that using `remainder.chars().count() == 1` may catch more cases (non-ASCII strings)
-                // but the main focus here is on " ", "-" and similar that are abundantly produced
-                // by `split_word_bounds()` in the Latin segmenter and mere `len()` performs better at that.
-                if remainder.len() == 1 {
-                    self.state = State::Exhausted;
-                    return Some(remainder);
-                }
-
-                match CAMEL_CASE_BOUNDARY_REGEX.find(remainder) {
-                    Some(mat) => {
-                        // By the nature of the regex, the match must contain at least two chars and this should never panic.
-                        let uppercase_letter_length =
-                            mat.as_str().chars().last().unwrap().len_utf8();
-                        let boundary = mat.end() - uppercase_letter_length;
-
+                match find_camel_case_boundary(remainder) {
+                    Some(boundary) => {
                         self.state = State::InProgress { remainder: &remainder[boundary..] };
                         Some(&remainder[..boundary])
                     }
@@ -64,6 +45,31 @@ impl<'t> Iterator for CamelCaseParts<'t> {
             }
         }
     }
+}
+
+fn find_camel_case_boundary(str: &str) -> Option<usize> {
+    // CamelCase boundary consists of at least 2 code-points. Avoid iterating over shorter strings.
+    // Note that using `remainder.chars().count() == 1` may catch more cases (non-ASCII strings)
+    // but the main focus here is on " ", "-" and similar that are abundantly produced
+    // by `split_word_bounds()` in the Latin segmenter and mere `len()` performs better at that.
+    if str.len() == 1 {
+        return None;
+    }
+
+    let mut last_char_was_lowercase = false;
+
+    str.find(|c: char| {
+        if c.is_mark_nonspacing() {
+            return false;
+        }
+
+        if last_char_was_lowercase && c.is_letter_uppercase() {
+            return true;
+        }
+
+        last_char_was_lowercase = c.is_letter_lowercase();
+        false
+    })
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


charabia/src/segmenter/camel_case.rs
```diff
@@ -66,6 +47,31 @@ impl<'t> Iterator for CamelCaseParts<'t> {
...
 
+fn find_camel_case_boundary(str: &str) -> Option<usize> {
+    // CamelCase boundary consists of at least 2 code-points. Avoid iterating over shorter strings.
+    // Note that using `remainder.chars().count() == 1` may catch more cases (non-ASCII strings)
+    // but the main focus here is on " ", "-" and similar that are abundantly produced
+    // by `split_word_bounds()` in the Latin segmenter and mere `len()` performs better at that.
+    if str.len() < 2 {
+        return None;
+    }
+
```
strohel, line 58: Omitting this optimization may now have negligible performance impact - hard to say without benchmark, but likely worth trying.

charabia/src/segmenter/camel_case.rs
```diff
@@ -66,6 +47,31 @@ impl<'t> Iterator for CamelCaseParts<'t> {
...
+    let mut last_char_was_lowercase = false;
+
+    str.find(|c: char| {
+        if c.is_mark_nonspacing() {
+            return false;
+        }
```
ManyTheFish, line 64: There will be 2 concurrent ways to check if a character is a nonspacing mark, see the normalizer below:
https://github.com/meilisearch/charabia/blob/main/charabia/src/normalizer/nonspacing_mark.rs#L36-L39

this isn't a big issue, but I'd prefer having 1 way to do it. I don't exclude the library you are using and it could probably replace the dictionary in the normalizer, I put below the related PR showing contributor motivations:
https://github.com/meilisearch/charabia/issues/137
https://github.com/meilisearch/charabia/pull/146

strohel (approved):
Continuing my for-the-fun reviews: LGTM! One little curiosity question.

ManyTheFish (changes requested):
Hello @goodhoko and @strohel,
I like the fact that we now avoid regex, so thanks for this.

Here are some suggestions:
1) what about wrapping `find_camel_case_boundary` in a [linear_group_by](https://docs.rs/slice-group-by/0.3.0/slice_group_by/) instead of `CamelCaseParts`, this could save you the job of splitting the text.
2) I've made a suggestion about `is_mark_nonspacing` because we use a similar function in an other part of Charabia

Thanks again!

charabia/src/segmenter/camel_case.rs
ManyTheFish: It's a good question,
I found a random study [trying to predict the frequency of words depending on their size](https://math.wvu.edu/~hdiamond/Math222F17/Sigurd_et_al-2004-Studia_Linguistica.pdf),
It seems that, in English, there is a bit more than 3% chance of encountering a word containing less than 2 characters. 🤔
I agree about the fact that this optimization seems superfluous. ☺️

goodhoko: Good point. I was hesitant to keep this in too, so I actually did the benchmark and saw an improvement of about 1-2%. That's very close to criterion's noise threshold but I was getting consistent numbers over multiple runs so I decided to keep it. Note that the numbers refer to performance of the whole latin segmenter. The impact on `split_camel_case_bounds()` alone would be higher but that's not of our interest here.

To see why optimizing for such a 'cornery' case yields any measurable improvement, one must look at the context where `split_camel_case_bounds()` [is used](https://github.com/meilisearch/charabia/blob/16665a6ced92d90bb0d5c5762ca7d23349f18113/charabia/src/segmenter/latin.rs#L20-L27):
```rust
fn segment_str<'o>(&self, s: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
    let lemmas = s
        .split_word_bounds()
        .flat_map(|lemma| lemma.split_inclusive('\''))
        .flat_map(|lemma| lemma.split_camel_case_bounds());

    Box::new(lemmas)
}
```

Here, `s` is the whole string we're segmenting. We first break it on word boundaries, then on apostrophes and then finally on camelCase boundaries. The important detail is that `split_word_bounds()` breaks apart the spacing between words. For instance, `"foo bar".split_word_bounds()` gives `["foo", " ", "bar"]. This optimization targets exactly these spaces and other one-character word separators. Intuitively, in longer texts (like used in the [benchmarks](https://github.com/meilisearch/charabia/blob/16665a6ced92d90bb0d5c5762ca7d23349f18113/charabia/benches/bench.rs#L11-L12)) the frequency of 1-char strings seen by `split_camel_case_bounds()` is close to 50%.*

Though, I have no strong feeling about this, so if you think that the tiny performance boost is not worth it I'll happily remove this.

---

* This raises the question: is there any performance to gain by shuffling the order in which we split on the various breakpoints. I just wanted to keep the scope small.

charabia/src/segmenter/camel_case.rs
goodhoko: Aha, good catch. Looking into the linked issue and PR it seems that the `finl_unicode` or other crates dedicated to finding Char's category were not considered. If this PR gets merged we'll need to check for more than `Mn` category anyway so I suppose we'll want to use a library.

I'm all up for measuring performance of the various crates and (the current HashSet implementation) to make an educated decision what approach to choose here. Though it would probably be best done in a separate PR?

ManyTheFish: @goodhoko, I agree with your approach of separating it into another PR.
Let's consider this as resolved

charabia/src/segmenter/camel_case.rs
ManyTheFish: Well, I may know why you are seeing improvement actually, I completely forgot the separators, if we take separators into account, we have much more lemmas containing 1 character than the 3% I could expect.
mmmmh... 🤔 honestly I don't have any strong point of view on this.
In fact, it's better to skip the process on separators, but, in another hand 1-2% gain is not so much compared to other optimization that could be done elsewhere in the Library.

ManyTheFish (approved):
Nice work, thanks to both of you,

I think the segmentation of `""` in `[]` shouldn't be an issue in the current implementation, however, I'll keep it in mind if an issue is raised on the Meilisearch side. 👍 

 bors merge

goodhoko:
@ManyTheFish 
> what about wrapping find_camel_case_boundary in a [linear_group_by](https://docs.rs/slice-group-by/0.3.0/slice_group_by/) instead of CamelCaseParts, this could save you the job of splitting the text.

So I tried my best to do this in [c9045e4](https://github.com/meilisearch/charabia/pull/196/commits/c9045e457077aba5d2e1aab53771304530f33414). Apart from the two consequences I mention in the commit message and which are IMO totally fine I see two problems:
1) I may be missing a way to write it more clearly but I'm not sure if the code is quite as understandable. There's much less of it for sure. I also feel that I'm now too deep into this to asses which version is clearer. What do *you* think?
2) As it's now that commit causes ~17% regression in performance.

curquiza:
bors merge

PR #205, opened 2023-04-08
Improve Arabic language segmentation

### LINKED ISSUE #133: Arabic script: Implement specialized Segmenter
Currently, the Arabic Script is segmented on whitespaces and punctuation.

### Drawback

Following the [dedicated discussion on Arabic Language support](https://github.com/meilisearch/product/discussions/139) and the linked issues, the agglutinative words are not segmented, for example [in this comments](https://github.com/meilisearch/charabia/issues/36#issuecomment-999935622):
>  the agglutinated word `الشجرة` => `The Tree` is a combination of `الـ` and `شجرة`
`الـ` is equivalent to `The` and it's always connected (not space separated) to the next word.

### Enhancement

We should find a specialized segmenter for the Arabic Script, or else, a dictionary to implement our own segmenter inspired by [the Thaï Segmenter](https://github.com/meilisearch/charabia/blob/main/src/segmenter/thai.rs).

------------------------
> Hey! 👋 

DrAliRagab:
# Pull Request

## Related issue
Fixes #133 

## What does this PR do?
Arabic text is segmented by word boundaries and by punctuation. We need a workaround to segment the Arabic text that starts with `ال` (the) because it is not segmented by word boundaries. One possible solution is to segment any word that starts with `ال` into two words. The `ال` and the rest of the word. with this solution, we will have `الشجرة` (the tree) segmented into `ال` (the) and `شجرة` (tree). and if we search for `شجرة` (tree) or `الشجرة` (thetree) we will find results. Some Arabic text starts with `ال` and not meant to be (the) like `البانيا` (Albania). In this case, we will have `ال` and `بانيا` segmented. and if we search for `البانيا` we will find results.

## PR checklist


DrAliRagab:
> Regarding segmenter, in Arabic language the starting `ال` is almost in 95% of words (I assume) is equal to `the`. As you can see here when I am searching for `الدوري` which means `the league` I got results. However when I search for `دوري` which means `league` I got no results. Screenshot 2023-04-06 194209 Screenshot 2023-04-06 194239
> 
> While in some other words like `البانيا` which means `Albania`, the starting `ال` is not equal to `the` and it's part of the word. but it will not make any significant difference in the search results. If I search for `بانيا` which equals to `bania` and get results it's not a big deal.
> 
> But if I search for `دوري` which equals to `league` and get no results, while there are many results for `الدوري` which equals to `the league`, it's a big deal.
> 
> So I think any word starts with `ال` should be segmented to two words, the one with `ال` and the one without `ال` and the search should be done on both of them.

Regarding the currently used segmenter for Arabic language, Any Arabic word that starts with `ال` are not segmented correctly. Because it's segmented based on space and punctuations.

To fix this; we can segment any word starts with `ال` into `ال` and the rest of the word. so when any user search for the word with `ال` or without `ال` related results will be shown.

While in some other words like `البانيا` which means `Albania`, the starting `ال` is not equal to `the` and it's part of the word. but it will not make any significant difference in the search results.
If I search for `البانيا` (Albania) or `بانيا` (bania), results will be shown.

After building Meilisearch with this segmenter and after rebuilding the indexes the issue was resolved.
As in this screenshot, when I searched for `دوري` (league) I got the results which fixes the previous issue.

Screenshot 2023-04-09 003319

charabia/src/segmenter/arabic.rs
```diff
@@ -0,0 +1,110 @@
...
+
+// All specialized segmenters only need to implement the method `segment_str` of the `Segmenter` trait.
+impl Segmenter for ArabicSegmenter {
+    fn segment_str<'o>(&self, to_segment: &'o str) -> Box<dyn Iterator<Item = &'o str> + 'o> {
+        // Create the iterator that will segment the provided text.
+        let segment_iterator = to_segment
+            // Split the text by word boundaries.
+            .split_word_bounds()
+            // Split the text by punctuation.
+            .flat_map(|lemma| lemma.split_inclusive(|c: char| c.is_ascii_punctuation()))
```
ManyTheFish, line 24:
```suggestion
```
Is it necessary, the function `split_word_bounds` should have already done the job, no? 🤔

ManyTheFish (changes requested):

ManyTheFish:
Hello @DrAliRagab, could you add some benchmarks for the Arabic Language?

You have to add 2 texts of different sizes, the script will do the rest!

https://github.com/meilisearch/charabia/blob/7349df3fd461ca95d7b71835d5816634e1d63274/charabia/benches/bench.rs#L7-L26

ManyTheFish (approved):
Nice work! Thank you for your contribution.

bors merge

PR #214, opened 2023-05-24
Transform classifier into normalizer

ManyTheFish:
This PR transforms the classifier into a normalizer easing the future implementation of the separator customization in Charabia

This PR also adds the `lossy_normalization` option allowing to activate or deactivate lossy normalization, this will be useful to normalize stop_words, custom separators, or custom words in order to be properly classified


dureuill:
Hello @ManyTheFish :wave:,

Thank you for the PR, I'll start reviewing it.

My understanding from our offline discussion is that this PR changes the behavior of stop words.

|Behavior|**Before** this PR|**After** this PR|
|---------|------------------|----------------|
|Case sensitivity (the != or == to The?)| Sensitive|***In*sensitive**|
|Diacritics sensitivity (the != or == to thé?) | Sensitive | Sensitive (**unchanged**)|

That is:

- Both before and after this PR, having `the` in the stop words does not make `thé` a stop word as well
- Before this PR, having `the` in the stop words would not make `The` a stop word as well, but after this PR **it does**.

If my understanding is correct, I think we should publicize the difference a little bit more so that product and documentation are aware of this change and what the new state is. A quick look through the documentation and specification indicates to me that the current case and diacritics sensitivity of stop words is not currently documented. I think it should probably be because as a user of Meilisearch I already wondered this about stop words.

So that we can do so, could you open an issue, to be fixed by this PR, and tag the documentation team + add the missing spec label on it? For the description of this issue, if you want to, you can reuse part of the tables and examples I gave at the beginning of this message.

I also have some additional questions: 

1. what is the current behavior of the stop words when a kanji is passed as a stop word? My expectation is that only the kanji is considered a stop word, not its various reading. As an example, adding 成犬 to the stop words does not add セイケン or せいけん.
2. what is the current behavior of the stop words when kana are passed as a stop word? My expectation is that only the kana that were passed are considered a stop word, such that adding せいけん to the stop words does not add セイケン, 成犬 or 聖剣.

charabia/src/normalizer/arabic.rs
```diff
@@ -202,6 +203,7 @@ mod test {
                     (2, 2),
                 ]),
                 script: Script::Arabic,
+                kind: TokenKind::Word,
```
dureuill, line 206: I'm not sure I understand why we need to specify the token kind here now?

charabia/src/tokenizer.rs
```diff
@@ -192,6 +194,19 @@ impl<'al, 'sw, A> TokenizerBuilder<'al, 'sw, A> {
         self
     }
 
+    /// Enable or disable the lossy normalization.
+    ///
+    /// A lossy normalization is a kind of normalization that could change the meaning in some way.
+    /// Removing diacritics is considered lossy, for instance, in French the word `maïs` (`corn`) will be normalized as `mais` (`but`) which change the meaning.
```
dureuill, line 200:
```suggestion
    /// Removing diacritics is considered lossy; for instance, in French the word `maïs` (`corn`) will be normalized as `mais` (`but`) which changes the meaning.
```

dureuill (commented):
I have a few questions and typos. Otherwise looks good, thanks 😊

charabia/src/normalizer/mod.rs
```diff
@@ -30,31 +33,38 @@ mod lowercase;
...
         Box::new(ChineseNormalizer),
         #[cfg(feature = "japanese-transliteration")]
         Box::new(JapaneseNormalizer),
         #[cfg(feature = "greek")]
         Box::new(GreekNormalizer),
-        Box::new(ControlCharNormalizer),
-        Box::new(QuoteNormalizer),
-        Box::new(NonspacingMarkNormalizer),
         Box::new(ArabicNormalizer),
+        Box::new(NonspacingMarkNormalizer),
```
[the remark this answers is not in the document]
ManyTheFish: No real change here, I just personnally prefer keeping the `NonspacingMarkNormalizer` at the last position.

charabia/src/normalizer/arabic.rs
ManyTheFish: There are 2 tests, one is a unit test over the current normalizer, and the other is a test running the whole normalization pipeline. In the second case, the classifier is now called and classifies each token forcing us to specify how the token should be classified.

dureuill: I'm feeling there should maybe be a comment in the code specifying that the classification is needed... On the other hand, I'm assuming that if we don't specify this then the test will fail?

If that is the case, we can resolve this.

charabia/src/normalizer/mod.rs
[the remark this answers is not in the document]
dureuill: OK

charabia/src/tokenizer.rs
```diff
@@ -79,42 +79,40 @@ pub trait Tokenize<'o, A: AsRef<[u8]>> {
...
-    normalizer_option: NormalizerOption,
+    normalizer_option: &'no NormalizerOption<'no>,
 }
 
-impl<'al, 'sw, A: AsRef<[u8]>> Tokenizer<'al, 'sw, A> {
+impl<'al, 'no> Tokenizer<'al, 'no> {
     /// Creates an Iterator over [`Token`]s.
     ///
     /// The provided text is segmented creating tokens,
     /// then tokens are normalized and classified.
```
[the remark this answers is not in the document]
dureuill: Yes, I understand this. However, classifying tokens is no longer the responsibility of this function:

- previously, this function was explicitly calling `classify_with_stop_words` to classify the tokens
- now, this function classifies the tokens implicitly by calling `normalize`, because the classifier is part of the list of normalizers. If we were to remove the classifier from the list of normalizers the comment would no longer be true even though we're not modifying the implementation of this function.

I'm guessing the semantic issue I'm having is that `normalize` and the list of normalizers now can contain a classifier. Maybe we should have used a term encompassing both normalization and classification (but this is probably just `tokenize`, so redundant here in the comment).

Maybe a solution could be to specify:

```suggestion
    /// then tokens are normalized and classified depending on the list of normalizers and classifiers in [`normalizer::NORMALIZERS`].
```

Then the relationship between what happens and this list is made explicit. What do you think?

ManyTheFish:
> Hello @ManyTheFish 👋,
> 
> Thank you for the PR, I'll start reviewing it.
> 
> My understanding from our offline discussion is that this PR changes the behavior of stop words.
> Behavior 	**Before** this PR 	**After** this PR
> Case sensitivity (the != or == to The?) 	Sensitive (`the` != `The`) 	**_In_sensitive** (`the` **==** `The`)
> Diacritics sensitivity (the != or == to thé?) 	Sensitive (`the` != `thé`) 	Sensitive (`the` != `thé`, **unchanged**)
> 
> That is:
> 
>     * Both before and after this PR, having `the` in the stop words does not make `thé` a stop word as well
> 
>     * Before this PR, having `the` in the stop words would not make `The` a stop word as well, but after this PR **it does**.
> 
> 
> If my understanding is correct, I think we should publicize the difference a little bit more so that product and documentation are aware of this change and what the new state is. A quick look through the documentation and specification indicates to me that the current case and diacritics sensitivity of stop words is not currently documented. I think it should probably be because as a user of Meilisearch I already wondered this about stop words.
> 
> So that we can do so, could you open an issue, to be fixed by this PR, and tag the documentation team + add the missing spec label on it? For the description of this issue, if you want to, you can reuse part of the tables and examples I gave at the beginning of this message.

mmmmh... You're probably right, I'm just wondering about the fact that we would commit to a design without ensuring that it's the best one. Let's "revert" the changes keeping the previous behaviors, and rethink it more globally later to commit to a final behavior to freeze in a spec.

> 
> I also have some additional questions:
> 
>     1. what is the current behavior of the stop words when a kanji is passed as a stop word? My expectation is that only the kanji is considered a stop word, not its various reading. As an example, adding 成犬 to the stop words does not add セイケン or せいけん.
> 
>     2. what is the current behavior of the stop words when kana are passed as a stop word? My expectation is that only the kana that were passed are considered a stop word, such that adding せいけん to the stop words does not add セイケン, 成犬 or 聖剣.

No Japanese normalization occurs before the classification, moreover, Charabia is not able to convert Kanji into Kana but only katakana into Hiragana.

dureuill:
Thanks for your answer many!

>  Let's "revert" the changes keeping the previous behaviors, and rethink it more globally later to commit to a final behavior to freeze in a spec.

I think we can keep this PR open for now, but yeah I'd rather we decided on a spec and behaviour before merging it (at least at the Meilisearch level). I think maybe this can be done quickly so that we don't spend too much time before merging this PR.

ManyTheFish:
@dureuill, I took into account your change requests on documentation and I move the Lowercase normalizer into the lossy normalizer imitating the previous behavior, this way we can merge this PR and decide later the order of the lists.

charabia/src/normalizer/arabic.rs
ManyTheFish: the test fails if this one is missing

dureuill (approved):
Alright looks like a good strategy! Let's move forward with this PR then :-)

ManyTheFish:
Thank you @dureuill,

bors merge

PR #215, opened 2023-05-29
Add an option to customize separators

ManyTheFish:
This PR adds a pre-segmenter that segments a text based on a separator list possibly given during the Tokenizer's building. Moreover, this separator's list is used by the classifier to categorize a lemma as a separator or not.

These changes highly impact the Hebrew and Latin tokenizers based on Unicode-segmentation:
- The Hebrew segmenter has been removed because the segmentation is completely done by the pre-segmenter
- The Latin segmenter only segments camel-cased words if the feature is activated.

About Hard separators (or context separators), because I didn't want to rely on deunicode to define a Hard separator and because I didn't manage to find an exhaustive list of "sentence/context separators" (in opposition to word separators), I personally created my own list of context separators based on several Unicode usages and conventions.


Kerollmops (approved):
Looks good to me 👍
bors merge

PR #231, opened 2023-07-18
Fix pre-segmenter when a string start by an uncategorized character

ManyTheFish:
When a field start with an uncategorized character like numbers these characters will be put in a separate group from the first categorized character which gives a different segmentation on the same word. For example, a field containing `Memory 96GB` will be split as `["Memory", " ", "96GB"]` and a query containing `96GB` will be split as `["96", "GB"]` because `96` is uncategorized where `GB` is `Latin`. Now we ensure to group every uncategorized character with the next script.

related to https://github.com/meilisearch/meilisearch/discussions/3924#discussioncomment-6468493


Kerollmops (approved):
Thank you for the fix @ManyTheFish!
Is this PR fixing https://github.com/meilisearch/meilisearch/discussions/3924#discussioncomment-6475892 or not?

ManyTheFish:
Hello @Kerollmops, at least a part of the bug comes from this issue.
I'm like 70% sure that will completely solve the bug

Kerollmops (approved):
bors merge

PR #234, opened 2023-08-21
Remap the char map when lowercasing strings

Kerollmops:
This PR is related to https://github.com/meilisearch/meilisearch-support/issues/1 and creates a test with a string making, producing a panic. The issue is due to the standard Rust lowercase function not recomputing the char map because it is stated that a lowercase doesn't change the size of the character when, in fact, [it does (stackoverflow)](https://stackoverflow.com/a/69060412/1941280), e.g. [`Ɦ` gives `ɦ`, 3 bytes to 2 bytes](https://mothereff.in/byte-counter#%EA%9E%AA%C9%A6).

### DIFF
--- a/charabia/src/normalizer/lowercase.rs
+++ b/charabia/src/normalizer/lowercase.rs
@@ -15,24 +15,18 @@ impl Normalizer for LowercaseNormalizer {
     fn normalize<'o>(&self, mut token: Token<'o>, _options: &NormalizerOption) -> Token<'o> {
         match token.char_map.take() {
             Some(char_map) => {
-                // Most of the time lowercasing a string will not change it's length
-                let lowercased = token.lemma.to_lowercase();
-                if token.lemma.len() == lowercased.len() {
-                    token.lemma = Cow::Owned(lowercased);
-                } else {
-                    let mut new_lemma = String::new(); // with_capacity?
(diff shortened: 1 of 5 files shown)


charabia/src/normalizer/lowercase.rs
```diff
@@ -9,10 +9,33 @@ use crate::Token;
...
+    // lowercasing characters cna change the characters length, so we need
+    // to make sure that the char mapping is correct and remap it if necessary.
+    // <https://github.com/meilisearch/charabia/pull/234>
     fn normalize<'o>(&self, mut token: Token<'o>, _options: &NormalizerOption) -> Token<'o> {
-        token.lemma = Cow::Owned(token.lemma().to_lowercase());
+        match token.char_map.take() {
+            Some(char_map) => {
+                // Most of the time lowercasing a string will not change it's length
+                let lowercased = token.lemma.to_lowercase();
+                if token.lemma.len() == lowercased.len() {
```
Kerollmops, line 20: Is it correct? One chunk of the string increases, and another may decrease to fake that nothing moved, but the chunk bounds moved, and bounds will be off. I fear that I'll need to remove this optimization...

Kerollmops:
As I am the only one available on the team, I will merge _by hand without approval 🆘

PR #240, opened 2023-09-10
Fuzz testing with `quickcheck` for normalizers, segmenters, tokenizer and classifier.

### LINKED ISSUE #235: Test this library against invalid and strange input
I would like that we use [the quickcheck or any equivalent library](https://github.com/BurntSushi/quickcheck) to test this crate. We must check that we can't panic when given strange input. We need to check:

 - The normalizers
 - The tokenizers
 - Random combinations of them
 - Normalizers with char maps

choznerol:
# Pull Request

## Related issue
Fixes #235 

## What does this PR do?

Close #235 by testing normalizers and segmenters with random tokens generated by `quickcheck`.

In addition, I also added a few property based test for `Tokenizer` (fb36a7308b290c40ff25f8df6f70340a75a0e00c) and `Classifier` (dcbf55cb120dece97f46c10c6c591a0b2af4904d). However, I'm not sure if they add much value to the existing tests, so I'd be happy to remove them if that's the case.

### Why the type alias `StaticToken`?

The [`Testable`](https://docs.rs/quickcheck/latest/quickcheck/trait.Testable.html) trait doesn't seem to support types with lifetime, even `'static`:
```rs
// error: implementation of `Testable` is not general enough
//    --> charabia/src/normalizer/mod.rs:374:5
//     |
// 374 |     #[quickcheck]
//     |     ^^^^^^^^^^^^^ implementation of `Testable` is not general enough
//     |
//     = note: `Testable` would have to be implemented for the type `for<'a> fn(token::Token<'a>)`
//     = note: ...but `Testable` is actually implemented for the type `fn(token::Token<'0>)`, for some specific lifetime `'0`
//     = note: this error originates in the attribute macro `quickcheck` (in Nightly builds, run with -Z macro-backtrace for more info)
#[quickcheck]
fn normalizer_should_not_panic(token: Token<'_>) {
    // ...
}
```
The type alias `StaticToken` was the easiest workaround I found.
Other more verbose workarounds like _`struct TokenWrapper { token: Token }`_ and _`struct OwnedToken { lemma: String }`_ could also work.

## PR checklist

### DIFF
--- a/charabia/src/token.rs
+++ b/charabia/src/token.rs
@@ -185,14 +185,24 @@ pub type StaticToken = Token<'static>;
 #[cfg(test)]
 impl Arbitrary for Token<'static> {
     fn arbitrary(g: &mut Gen) -> Self {
+        let lemma = String::arbitrary(g);
+
+        let bytes_count = lemma.len();
+        let byte_start = usize::arbitrary(g).saturating_sub(bytes_count);
+        let byte_end = byte_start + bytes_count;
+
+        let chars_count = lemma.chars().count();
+        let char_start = usize::arbitrary(g).saturating_sub(chars_count);
+        let char_end = char_start + chars_count;
+
         Token {
             kind: TokenKind::arbitrary(g),
             lemma: Cow::Owned(String::arbitrary(g)),
-            char_start: usize::arbitrary(g),
-            char_end: usize::arbitrary(g),
-            byte_start: usize::arbitrary(g),
-            byte_end: usize::arbitrary(g),
             char_map: Option::arbitrary(g),
+            char_start,
+            char_end,
+            byte_start,
+            byte_end,
             script: Script::arbitrary(g),
             language: Option::arbitrary(g),
         }
--- a/charabia/src/token.rs
+++ b/charabia/src/token.rs
@@ -198,11 +198,11 @@ impl Arbitrary for Token<'static> {
         Token {
             kind: TokenKind::arbitrary(g),
             lemma: Cow::Owned(String::arbitrary(g)),
-            char_map: Option::arbitrary(g),
             char_start,
             char_end,
             byte_start,
             byte_end,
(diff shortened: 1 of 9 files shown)


charabia/src/token.rs
```diff
@@ -161,3 +177,34 @@ impl Token<'_> {
...
+        let char_end = char_start + chars_count;
+
+        Token {
+            kind: TokenKind::arbitrary(g),
+            lemma: Cow::Owned(String::arbitrary(g)),
+            char_start,
+            char_end,
+            byte_start,
+            byte_end,
+            char_map: None,
```
choznerol, line 205: Using a random incorrect `char_map` here can trigger a lot of out of UTF-8 boundary panic, but I assume that's not something we are interested in.
```rs
            char_map: Option::arbitrary(g),
```

ManyTheFish: Hello @choznerol,

You could run a [normalization over a str](https://docs.rs/charabia/0.8.3/charabia/normalizer/trait.Normalize.html) activating the char_map creation, then check if the method [Token::original_lengths](https://docs.rs/charabia/0.8.3/charabia/struct.Token.html#method.original_lengths) doesn't panic if it's called for any length.

choznerol: > You could run a [normalization over a str](https://docs.rs/charabia/0.8.3/charabia/normalizer/trait.Normalize.html) activating the char_map creation, then check if the method [Token::original_lengths](https://docs.rs/charabia/0.8.3/charabia/struct.Token.html#method.original_lengths) doesn't panic if it's called for any length.

Thanks! Fixed in 2e5e2c3

ManyTheFish (approved):
Hello @choznerol sorry for the delay,
LGTM,
Thank you for your contribution,

bors merge

choznerol:
No worry! Thanks for the PR review 😄

PR #280, opened 2024-04-15
Add `\t` as recognized separator

Gusted:
Currently `\t` isn't seen as an recognized separator. This was causing issues for meilisearch, when it was trying to search on a keyword (fuzzy or exact match) and in the document the keyword was present but the character before the keyword was an `\t` charabia would create a token that was `\t<keyword>` which in turn led to meilisearch returning the document as part of the search but not returning the positions of matches (`_matchesPosition` field).

The actual reproducer for this bug was code files of the Linux kernel (such as `fs/ext4/readpage.c`) which uses tabs for indentation and searching for keywords like `while` would usually be 'prefixed' by an tab causing the described issue. Making `\t` a separator fixed this issue.


ManyTheFish (approved):
Yes Indeed, you are right,
Thank you for your contribution,
bors merge

PR #281, opened 2024-04-17
Fix char boundary panic

ManyTheFish:
Filter empty tokens before inserting them in the AhoCorasick automaton, avoiding a char boundary panic

## Related issue
Fixes https://github.com/meilisearch/meilisearch/issues/4574


curquiza:
@Kerollmops because until it's not fully merged in Meilisearch, it does not completely fix the issue. I did not want this PR to close the MEilisearch issue.

PR #282, opened 2024-04-17
Make the pinyin-normalization optional

ManyTheFish:
- split the Chinese feature flag in order to deactivate a part of the pipeline
- remove pinyin normalization from the default pipeline

### DIFF
--- a/charabia/Cargo.toml
+++ b/charabia/Cargo.toml
@@ -40,7 +40,10 @@ zerovec = "0.10.1"
 default = ["chinese", "hebrew", "japanese", "thai", "korean", "greek", "latin-camelcase", "latin-snakecase", "khmer", "vietnamese"]
 
 # allow chinese specialized tokenization
-chinese = ["dep:pinyin", "dep:jieba-rs"]
+chinese = ["chinese-segmentation", "chinese-normalization"]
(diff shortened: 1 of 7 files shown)


charabia/Cargo.toml
```diff
@@ -40,7 +40,10 @@ zerovec = "0.10.1"
 default = ["chinese", "hebrew", "japanese", "thai", "korean", "greek", "latin-camelcase", "latin-snakecase", "khmer", "vietnamese"]
 
 # allow chinese specialized tokenization
-chinese = ["dep:pinyin", "dep:jieba-rs"]
+chinese = ["chinese-segmentation", "chinese-normalization"]
+chinese-segmentation = ["dep:jieba-rs"]
+chinese-normalization = []
+chinese-normalization-pinyin = ["dep:pinyin", "chinese-normalization"]
```
Kerollmops, line 46: Will you reflect that on Meilisearch, too?

Kerollmops (approved):
I didn't understand everything, but I agree with you!
bors merge

charabia/Cargo.toml
ManyTheFish: no, I don't think so, 
everybody complains about this normalizer but I keep it in case another user of charabia wants to keep it

PR #299, opened 2024-07-10
Simplify lang detection

ManyTheFish:
- Change the language `allow_list` from a map of script->language to an array of allowed languages
- Allow to dynamically change the language allow list when tokenizing text


charabia/src/detection/script_language.rs
```diff
@@ -361,12 +353,12 @@ mod test {
...
-        assert_eq!(Language::Jpn.name(), "jpn");
-        assert_eq!(Language::from_name("jpn"), Language::Jpn);
-        assert_eq!(Language::Cmn.name(), "cmn");
-        assert_eq!(Language::from_name("cmn"), Language::Cmn);
+        assert_eq!(Language::Eng.code(), "eng");
+        assert_eq!(Language::from_code("eng"), Some(Language::Eng));
+        assert_eq!(Language::Jpn.code(), "jpn");
+        assert_eq!(Language::from_code("jpn"), Some(Language::Jpn));
+        assert_eq!(Language::Cmn.code(), "cmn");
+        assert_eq!(Language::from_code("cmn"), Some(Language::Cmn));
```
dureuill, line 361: not sure if this is relevant, but I thought we were using https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes. I would have expected just `en` for English, `ja` for Japanese, etc.

Looks like we're not using set 1, but the other sets? Is set 1 usable?

charabia/src/tokenizer.rs
```diff
@@ -354,6 +385,15 @@ impl<'tb, A: AsRef<[u8]>> TokenizerBuilder<'tb, A> {
             segmenter_option: Cow::Owned(self.segmenter_option),
         }
     }
+
+    /// Build the tokenizer options consumming self.
```
dureuill, line 389:
```suggestion
    /// Build the tokenizer options consuming self.
```

charabia/src/tokenizer.rs
```diff
@@ -354,6 +385,15 @@ impl<'tb, A: AsRef<[u8]>> TokenizerBuilder<'tb, A> {
...
+    ///
+    /// This method allows to drop the tokenizer builder without having to drop the options themselves.
+    pub fn into_options(mut self) -> (NormalizerOption<'tb>, SegmenterOption<'tb>) {
+        drop(self.build());
```
dureuill, line 393: unclear to me why the drop is needed.

Does `self.build()` mutate `self.normalizer_option` and `self.segmenter_option`?

charabia/src/detection/script_language.rs
ManyTheFish: These codes are inherited from [whatlang, which follows the ISO639-3](https://docs.rs/whatlang/latest/whatlang/enum.Lang.html), and the bigrams codes come from the  ISO639-1, which is an anterior version.
Is there any issue with using a 3-gram code?

dureuill: This is about user expection. Maybe that's just me, but as a user, I would expect my language codes to be "EN" or "en", not "ENG"

charabia/src/tokenizer.rs
```diff
@@ -100,32 +99,64 @@ impl Tokenize<'_> for &str {
 /// See [`TokenizerBuilder`] to know how to build a [`Tokenizer`].
 #[derive(Debug)]
 pub struct Tokenizer<'tb> {
-    segmenter_option: Cow<'tb, SegmenterOption<'tb>>,
-    normalizer_option: Cow<'tb, NormalizerOption<'tb>>,
+    pub segmenter_option: Cow<'tb, SegmenterOption<'tb>>,
+    pub normalizer_option: Cow<'tb, NormalizerOption<'tb>>,
```
ManyTheFish, line 103: do not publish inner anymore

ManyTheFish, line 103:
```suggestion
    segmenter_option: Cow<'tb, SegmenterOption<'tb>>,
    normalizer_option: Cow<'tb, NormalizerOption<'tb>>,
```

dureuill (approved):
bors merge

PR #303, opened 2024-08-09
feat: Adds German compound words decomposition with new segmenter

luflow:
# Pull Request

## What does this PR do?
- Adds first version of decomposition for german compound words based on a dictionary (based on https://github.com/uschindler/german-decompounder/)
- Adds benchmark with german sentences

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?


luflow:
I assume this could be a very expensive algorithm because all word lengths are checked against the dict?

Not sure if there is a better solution, but at least a first version for compound words :)

luflow:
Also another open question: can we even use the dictionary?

The orignal author has it under GNU GPL
https://github.com/uschindler/german-decompounder/blob/master/NOTICE.txt

luflow:
@curquiza @ManyTheFish fixed the fmt and clippy issues

charabia/src/segmenter/mod.rs
```diff
@@ -6,6 +6,7 @@ pub use arabic::ArabicSegmenter;
 #[cfg(feature = "chinese-segmentation")]
 pub use chinese::ChineseSegmenter;
 use either::Either;
+pub use german::GermanSegmenter;
```
ManyTheFish, line 9:
```suggestion
#[cfg(feature = "german-segmentation")]
pub use german::GermanSegmenter;
```

charabia/src/segmenter/mod.rs
```diff
@@ -25,6 +26,7 @@ use crate::token::Token;
 mod arabic;
 #[cfg(feature = "chinese-segmentation")]
 mod chinese;
+mod german;
```
ManyTheFish, line 29:
```suggestion
#[cfg(feature = "german-segmentation")]
mod german;
```

charabia/src/segmenter/mod.rs
```diff
@@ -72,6 +74,8 @@ pub static SEGMENTERS: Lazy<SegmenterMap> = Lazy::new(|| {
         ((Script::Khmer, Some(Language::Khm)), Box::new(KhmerSegmenter) as Box<dyn Segmenter>),
         // arabic segmenter
         ((Script::Arabic, Some(Language::Ara)), Box::new(ArabicSegmenter) as Box<dyn Segmenter>),
+        // german segmenter
+        ((Script::Latin, Some(Language::Deu)), Box::new(GermanSegmenter) as Box<dyn Segmenter>),
```
ManyTheFish, line 78:
```suggestion
        // german segmenter
        #[cfg(feature = "german-segmentation")]
        ((Script::Latin, Some(Language::Deu)), Box::new(GermanSegmenter) as Box<dyn Segmenter>),
```

ManyTheFish (changes requested):
Hello @luflow,

Could you add a feature flag on your implementation as I suggested, please? Then add it as a default feature in the `Cargo.toml` file.

In terms of implementation, you chose to rely on an HashSet to split your words, but I don't think it's the best approach.
I highly suggest using an `FstSegmenter` like in the [Thai tokenizer](https://github.com/meilisearch/charabia/blob/main/charabia/src/segmenter/thai.rs), it's a bit complex to build but way more efficient in time and space, or you could use an [AhoCorasick automaton](https://github.com/meilisearch/charabia/blob/9854134f64075bf763adda4672e775eaa720a330/charabia/src/tokenizer.rs#L348-L351) using the `LeftmostLongest` match kind.

Sorry for the delays!
Let me know if you have a question

luflow:
Hi @ManyTheFish!

Do you have any instructions to build the fst file? I could not find any material online - especially because FST is also used in other contexts like R but does something totally different 🤣

Otherwise the leftmostmatch functionality also works with a word dictionary if i understand it correctly?

ManyTheFish:
> Do you have any instructions to build the fst file? I could not find any material online - especially because FST is also used in other contexts like R but does something totally different 🤣

You can use the [CLI fst-bin](https://crates.io/crates/fst-bin) to build your dictionary from a source file. 😄 

> Otherwise the leftmostmatch functionality also works with a word dictionary if i understand it correctly?

Yes you can build it from an iterator over str, so it's convenient

charabia/src/segmenter/mod.rs
luflow: Done :)

charabia/src/segmenter/mod.rs
luflow: Done :)

charabia/src/segmenter/mod.rs
luflow: Done :)

luflow:
@ManyTheFish I extended the `FstSegmenter` with two options to also be able to handle a min lemma length and being able to hinder the segmenter from spitting out single letters. That keeps my dictionary even smaller and may be also useful for other languages later?

The dictionary is now also transformed into an FST file.

Let me know what you think :)

luflow:
@ManyTheFish dud you find time yet to look over the changes? Do you need anything else from my side? :)

ManyTheFish (approved):
Hello @luflow,
sorry for the delay, LGTM!

bors merge

charabia/src/segmenter/utils.rs
```diff
@@ -23,15 +29,40 @@ impl<'fst> FstSegmenter<'fst> {
...
+                    length = to_segment.len();
+                }
+            }
+
+            // ensure the length is a valid character boundary
+            length = to_segment
+                .char_indices()
+                .nth(length)
+                .map(|(idx, _)| idx)
+                .unwrap_or(to_segment.len());
```
ManyTheFish, line 64: Because the length is a byte length, we can't use nth because it count characters and not bytes:
```suggestion
            // ensure the length is a valid character boundary
            length = to_segment
                .char_indices()
                .find(|(idx, _)| idx >= length)
                .map(|(idx, _)| idx)
                .unwrap_or(to_segment.len());
```

luflow:
@ManyTheFish ok applied suggestion :)

ManyTheFish:
Hello @luflow,

the test and clippy are not happy,

could you ensure that:
- `cargo clippy`
- `cargo test`

work on your machine please?

I'll merge as soon as the tests pass 😃

luflow:
@ManyTheFish done 👍🏻

ManyTheFish (approved):
Nice!

Thank you for the contribution!

bors merge

Issue #322, opened 2025-01-07
Allow loading the lindera dictionaries at runtime instead of compiling them in


slatian:
Currently charabia uses lindera with the option to download the dictionary (~150MB) at build time from a CDN at build time which causes problems on multiple occasions, bot in terms of binary size and in terms of running a pretty complicated build script, that does what many build pipelines intentionally forbid (#265).

My concrete feature request would be some additional code that if enabled via feature flag allows loading the required dictionaries at runtime from external files, that can be managed decoupled from the software. (especially helpful when for example a backend tool and server are split up, but both need charabia + dictionaries as a dependencies)

Related: [Meilisearch Discussion #800: [Charabia] Add possibility to load dictionaries from external files](https://github.com/orgs/meilisearch/discussions/800)

Ai80-8:
Delete it all

PR #326, opened 2025-01-24
Adds support for the use of external dictionaries for segmenters backed by lindera

PedroTurik:
# Pull Request

## Related issue
Fixes [#322](https://github.com/meilisearch/charabia/issues/322)

## What does this PR do?
- This PR adds the features korean-segmentation-external and japanese-segmentation-external, that allow the user to decouple the download of the japanese and korean dictionaries from the compilation process, and configure the path to already downloaded lindera compatible dictionaries, with the MEILISEARCH_JAPANESE_EXTERNAL_DICTIONARY and MEILISEARCH_KOREAN_EXTERNAL_DICTIONARY env vars.

this PR is **not finished.** Since we cant control which dict the user will use, we cant be sure about the segmentation process, so activating the features will disable segmentation tests for now. Another thing worth mentioning is that for lindera to use an external dict, you need generate it through the lindera CLI. The process isnt exactly obvious and needs to be documented. Its described [here](https://github.com/lindera/lindera/issues/436)

### DIFF
--- a/charabia/Cargo.toml
+++ b/charabia/Cargo.toml
@@ -46,10 +46,13 @@ hebrew = []
 japanese = ["japanese-segmentation-unidic", "japanese-transliteration"]
 japanese-segmentation-ipadic = ["lindera/ipadic", "lindera/compress"]
 japanese-segmentation-unidic = ["lindera/unidic", "lindera/compress"]
+japanese-segmentation-external = ["lindera/compress"]
 japanese-transliteration = ["dep:wana_kana"]
 
 # allow korean specialized tokenization
-korean = ["lindera/ko-dic", "lindera/compress"]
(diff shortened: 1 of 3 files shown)


slatian:
Thanks for working on this!

If writing documentation is currently holding this back, I'd help out with that.

PedroTurik:
Hey @slatian

Yes, you are correct. Where do you think I should document the process to use this feature?

Also, thanks @ManyTheFish. I will fix the CI error and open the PR for review as you said

slatian:
> Where do you think I should document the process to use this feature?

As a user of the crate I'd expect the documentation for this with the rest of the documentation on docs.rs -> Document it in a module.

Depending on the length, either a `doc` module or a section beneath the "Build features" section in the main module.

PR #380, opened 2026-07-21
Make allow list API more flexible

RemiBardon:
Hey!

When implementing https://github.com/valeriansaliou/sonic/issues/375 I got slowed down because the `charabia` API was too restrictive regarding language allow lists. I had a `Option<charabia::Language>`, and your API forced me to give a `&[charabia::Language]`. I couldn’t use [`std::option::Option::as_slice`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_slice) because of lifetime issues, and I didn’t want it to impact all of my code, so I decided to fix the issue upstream.

Conceptually, what you want is to iterate over `charabia::Language`s. I made it so your API now accepts any `IntoIterator<Item = &Language> + Copy`, which adds support for `&Option<T>`, `&Box<[T]>`, etc. without breaking the current `&[T]` API.

I was hoping it could be a non-breaking change, but unfortunately the signature of `pub trait Tokenize` changed slightly which will break implementations. The fix is trivial though, it only consists of changing two `<'_, '_, '_, '_>` into `<'_, '_, '_, &'_ [Language]>`.

Although I doubt it’s used by a lot of people, as `&str` is likely the only type anyone expects to implement this helper trait, I wouldn’t be surprised if you chose to keep the API stable and dismiss this PR.

## Summary by CodeRabbit

* **New Features**
  * Added flexible language allow-list support across detection, segmentation, normalization, and tokenization.
  * Allow-lists can now be provided through multiple compatible collection types.
  * Existing convenience methods continue to use the default language list behavior.

### DIFF
--- a/charabia/src/detection/mod.rs
+++ b/charabia/src/detection/mod.rs
@@ -6,61 +6,75 @@ use whatlang::Detector;
 mod chars;
 mod script_language;
 
-pub struct StrDetection<'o, 'al> {
+pub struct StrDetection<'o, AllowList> {
     inner: &'o str,
     pub script: Option<Script>,
     pub language: Option<Language>,
-    allow_list: Option<&'al [Language]>,
+    allow_list: Option<AllowList>,
 }
 
-impl<'o, 'al> StrDetection<'o, 'al> {
-    pub fn new(inner: &'o str, allow_list: Option<&'al [Language]>) -> Self {
+impl<'o, AllowList> StrDetection<'o, AllowList> {
+    pub fn new(inner: &'o str, allow_list: Option<AllowList>) -> Self {
         Self { inner, script: None, language: None, allow_list }
     }
 
     pub fn script(&mut self) -> Script {
         let inner = self.inner;
-        *self.script.get_or_insert_with(|| Self::detect_script(inner))
+        *self.script.get_or_insert_with(|| detect_script(inner))
     }
 
-    pub fn language(&mut self) -> Option<Language> {
+    pub fn language<'lang, Lang>(&mut self) -> Option<Language>
+    where
+        AllowList: IntoIterator<Item = Lang> + Copy,
+        Lang: std::borrow::Borrow<Language>,
+    {
         let inner = self.inner;
         self.language = match self.language.take() {
             Some(lang) => Some(lang),
             None => match self.allow_list {
-                Some([unique_language]) => Some(*unique_language),
-                None if Self::detect_script(inner) == Script::Latin => None,
-                _otherwise => Self::detect_lang(inner, self.allow_list),
+                Some(ref allow_list) => {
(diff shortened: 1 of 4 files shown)


charabia/src/detection/mod.rs
```diff
@@ -6,61 +6,74 @@ use whatlang::Detector;
...
+                Some(ref allow_list) => {
+                    let mut iter = allow_list.into_iter();
+
+                    match iter.next() {
+                        Some(&lang) if iter.next().is_none() => Some(lang),
+                        _ => detect_lang(inner, Some(allow_list.into_iter())),
+                    }
+                }
+                None if detect_script(inner) == Script::Latin => None,
+                None => detect_lang::<<AllowList as IntoIterator>::IntoIter>(inner, None),
```
RemiBardon: This is the most error-prone part. The rest is just type changes.

Both versions should be equivalent, but I may have missed something.

charabia/src/tokenizer.rs
```diff
@@ -124,12 +130,39 @@ impl Tokenizer<'_> {
...
+    /// tokenizer.tokenize_with_allow_list(orig, Some(&[Language::Eng]));
+    ///
+    /// // tokenize by passing an optional.
+    /// let hint: Option<Language> = Some(Language::Eng);
+    /// tokenizer.tokenize_with_allow_list(orig, Some(&hint));
+    ///
+    /// // tokenize by passing a boxed slice.
+    /// let allow_list: Box<[Language]> = Box::from([Language::Eng]);
+    /// tokenizer.tokenize_with_allow_list(orig, Some(&allow_list));
+    /// ```
```
RemiBardon, line 160: And here is an example of the new API’s flexibility.

RemiBardon (commented):
Just a few comments

RemiBardon:
Oh no, this would conflict with Baptiste’s #378 🥲 You can merge his first and I’ll update mine.

RemiBardon:
Wait I’ll try to remove the `Copy` requirement, it’s still annoying me.

RemiBardon:
Actually it’s not `Copy` that was annoying me, but rather not being able to pass an `IntoIterator<Item = Language>`. I fixed that. Now one can pass `&[T]`, `Option<T>`, `Option<&T>`, `Box<[T]>`, `&Box<[T]>`, etc.

RemiBardon:
I’ll also make `TokenizerBuilder` more flexible while I’m at it, I realized that’s what I really needed as `segment_with_option` doesn’t classify segments as separators etc…

Edit: There was already `tokenize_with_allow_list` but it’s better if I also fix higher up in the builder itself.

RemiBardon:
I had a try and it would make too many API breaking changes. I decided to not change `TokenizerBuilder` as `Tokenizer` can already be used to achieve my goal.

PR #383, opened 2026-08-29
Normalize final sigma in every position, not only at the end

jaideeppyne:
`ς` and `σ` both case-fold to U+03C3 (CaseFolding.txt, status C), so they should always reach the same index term. The normalizer used `strip_suffix('ς')`, which only rewrites a sigma in last position.

That holds for well-formed modern Greek, where `ς` only appears word-finally. It breaks as soon as anything follows the sigma inside the token, because the sigma is then no longer the last char:

```
"κόσμος"          -> "κοσμοσ"
"κόσμος" + U+0301 -> "κοσμος"
```

`NonspacingMarkNormalizer` runs after this one and strips the mark, so those two forms differ by one character in the index and stop matching. Same for U+FE0F and U+200D. Combining dot below (U+0323) is routine in epigraphic and papyrological Greek, and stray combining marks are common in OCR and PDF-extracted text.

#179 states the rule with no positional qualifier ("ς = σ"), so I fixed every position instead of only the trailing-mark case.

I replaced the suffix check with a `find` plus a rewrite from the first `ς`. Both characters are 2 bytes in UTF-8, so the existing char_map skip stays valid.

How I found it: I diffed charabia's normalized output against Unicode full case folding for the sigma class, over 896 case-fold-equivalent Greek word pairs. 503 failed before the change, 0 after. The two new tests fail on current main and pass with the patch.

This is a hot path so I benchmarked it, 88k Greek tokens, best of 10, interleaved rounds: 21.10ms before, 20.90ms after. Writing it as `contains` + `replace` did cost about 2%, which is why it is a single `find`.

AI disclosure: I used Claude Code to build the differential harness and draft the patch. I reviewed, ran and benchmarked everything myself.

## Summary by CodeRabbit

- **Bug Fixes**
  - Greek lemma normalization now consistently converts final sigma (`ς`) to standard sigma (`σ`) wherever it appears.
  - Improved handling of final sigma followed by combining marks, ensuring equivalent marked and unmarked forms normalize identically.

- **Tests**
  - Added coverage for sigma normalization at the beginning, middle, and end of words.
  - Added coverage for sigma followed by a nonspacing combining mark, including comparisons between marked and unmarked forms.

### DIFF
--- a/charabia/src/normalizer/greek.rs
+++ b/charabia/src/normalizer/greek.rs
@@ -12,11 +12,19 @@ impl Normalizer for GreekNormalizer {
     // converting  "ς" to "σ" doesn't change the characters length,
     // so the `normalize` method is overloaded to skip the useless char_map computing.
     fn normalize<'o>(&self, mut token: Token<'o>, _options: &NormalizerOption) -> Token<'o> {
-        if let Some(index) = token.lemma.find('ς') {
-            let mut lemma = String::with_capacity(token.lemma.len());
-            lemma.push_str(&token.lemma[..index]);
-            lemma.extend(token.lemma[index..].chars().map(|c| if c == 'ς' { 'σ' } else { c }));
-            token.lemma = Cow::Owned(lemma)
+        if token.lemma.contains('ς') {
+            match token.lemma {
+                Cow::Borrowed(lemma) => token.lemma = Cow::Owned(lemma.replace('ς', "σ")),
+                Cow::Owned(mut lemma) => {
+                    let mut start = 0;
+                    while let Some(index) = lemma[start..].find('ς') {
+                        start += index;
+                        lemma.replace_range(start..start + 'ς'.len_utf8(), "σ");
+                        start += 'σ'.len_utf8();
+                    }
+                    token.lemma = Cow::Owned(lemma);
+                }
+            }
         }
 
         token
--- a/charabia/src/normalizer/greek.rs
+++ b/charabia/src/normalizer/greek.rs
@@ -4,7 +4,7 @@ use super::{Normalizer, NormalizerOption};
 use crate::{Script, Token};
 
 /// Normalize Greek characters by:
-/// 1. convert final sigma into ordinary sigma
+/// 1. convert final sigma (`ς`) into ordinary sigma (`σ`)
 ///
 pub struct GreekNormalizer;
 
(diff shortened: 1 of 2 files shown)


charabia/src/normalizer/greek.rs
```diff
@@ -4,16 +4,19 @@ use super::{Normalizer, NormalizerOption};
...
     // converting  "ς" to "σ" doesn't change the characters length,
     // so the `normalize` method is overloaded to skip the useless char_map computing.
     fn normalize<'o>(&self, mut token: Token<'o>, _options: &NormalizerOption) -> Token<'o> {
-        if let Some(prefix) = token.lemma.strip_suffix('ς') {
-            token.lemma = Cow::Owned([prefix, "σ"].concat())
+        if let Some(index) = token.lemma.find('ς') {
+            let mut lemma = String::with_capacity(token.lemma.len());
+            lemma.push_str(&token.lemma[..index]);
+            lemma.extend(token.lemma[index..].chars().map(|c| if c == 'ς' { 'σ' } else { c }));
+            token.lemma = Cow::Owned(lemma)
```
ManyTheFish: Find will stop on the first occurrence; if there are multiple occurrences in your token, only the first one will be replaced.

Moreover, you reallocate the full String even if it has already been allocated in a Cow::Owned. It might be wiser to replace the character.

ManyTheFish (changes requested):

charabia/src/normalizer/greek.rs
jaideeppyne: Addressed in 285e9ec. The normalizer now loops over every `ς` occurrence instead of stopping at the first one. It also preserves the existing owned allocation path by using `replace_range()` in place for `Cow::Owned`, while borrowed lemmas still take the normal `replace()` path to create the owned normalized value.

Validation: `cargo test -p charabia normalizer::greek` passed (5 tests).

jaideeppyne:
The requested allocation-preserving normalization change is in `285e9ec`: all sigma occurrences are replaced, and owned `Cow` values are updated in place with `replace_range`. The inline thread is resolved; targeted tests and `cargo fmt --check` pass.

ManyTheFish:
`cargo clippy` doesn't pass. Otherwise, this PR is mergeable

jaideeppyne:
Fixed in `9b66767`. Rust 1.98 introduced `clippy::chunks_exact_to_as_chunks`, which was failing in the existing nonspacing-mark dictionary loader under `--deny warnings`. I switched it to `as_chunks::<4>()`. `cargo +1.98.0 clippy --all-targets -- --deny warnings`, `cargo fmt --all -- --check`, and the five Greek normalizer tests all pass locally.

jaideeppyne:
@ManyTheFish The clippy failure on 1.98 (`chunks_exact_to_as_chunks` in the nonspacing-mark loader) is fixed in 9b66767, and the earlier allocation-preserving sigma change is in 285e9ec. The inline thread is resolved. This should be ready for another look.

jaideeppyne:
@ManyTheFish gentle bump whenever you have a moment — clippy on 1.98 is fixed in 9b66767 and the earlier review notes are addressed. Happy to adjust further if anything else stands out.

